PR #55, opened 2023-11-11
Add Git SHA api endpoint.

trilleplay:


oliverlevay (approved):
should work! unclear if this will work when deployed, but we will find out

danieladugyan:
> should work! unclear if this will work when deployed, but we will find out

I'm pretty sure it won't unfortunately. Maybe something along these lines? https://stackoverflow.com/questions/70436753/how-to-add-commit-hash-into-reactjs-vite-config-js

oliverlevay (changes requested):
I agree with Daniel

danieladugyan:
This PR has been stale for a while, so I'm closing it for now. Anyone can feel free to reopen this if they want to resume work on this feature.

PR #57, opened 2023-11-11
First fix for meta head titles

alfredgrip:
First fix for meta head titles, needs to respect preferred language later.

`<svelte:head>` should be put in `+page.svelte` or component files instead on `+layout.svelte` files.

Minor TS changes with types.


oliverlevay (approved):
LGTM 👍

PR #62, opened 2023-11-11
Create new markdown pages and grant permission set for those.

trilleplay:


danieladugyan (approved):
Looks good, just take a look at why CI fails

PR #100, opened 2023-11-15
Add pre commit hook

alfredgrip:
This adds a pre commit hook that formats and lints staged code to keep consistent code. Have not tried it extensively but it seems to work good.

Perhaps this could be set up in CI also/instead?


danieladugyan:
A pre-push hook has been merged now. It's quite slow and mirrors our CI. Takes about 30s for linting, checking formatting and svelte-check (mainly compilation). I think it's a good middle-ground, but we could consider adding a more lightweight pre-commit hook.

PR #104, opened 2023-11-16
[Suggestion & Implementation] Adding frosted glass look to Navbar

stagrim:
image

Feel free to adjust the effect [here](https://github.com/stagrim/web/blob/6f1796e75d18303144c4c0b685ed81569529f444/src/routes/NavBar.svelte#L182-L185) to fit it better.


src/routes/+layout.svelte
```diff
@@ -17,9 +17,9 @@
 </script>
 
 <NavBar>
-  <div
-    class="h-[calc(100vh-9rem)] flex-col overflow-auto pb-8 accent-primary md:h-[calc(100vh-5rem)] [&>*]:flex-1"
-  >
+  <div class="overflow-auto bg-base-200 accent-primary" style="height: 4.4rem"></div>
```
danieladugyan, line 20: We try and stick exclusively to Tailwind classes. In this case, something like `h-16` would be equivalent to the style attribute used.

src/routes/NavBar.svelte
```diff
@@ -176,3 +176,12 @@
...
 </div>
+
+<style>
+  .glass {
+    background: hsl(var(--b2) / 0.6);
+    /* box-shadow: 0 4px 30px rgba(0, 0, 0, 0.1); */
+    backdrop-filter: blur(10.5px);
+    -webkit-backdrop-filter: blur(10.5px);
+  }
+</style>
```
danieladugyan, line 187: Similarly, all these could be written as Tailwind classes (also removing the commented out line): `bg-<some color from our theme>` (see tailwind.config.js or just rely on Tailwind IntelliSense) and `filter backdrop-blur`.

package.json
```diff
@@ -64,6 +64,7 @@
     "isomorphic-dompurify": "^1.9.0",
     "marked": "^9.0.3",
     "minio": "^7.1.3",
+    "nopt": "^7.2.0",
```
danieladugyan, line 67: Do we need this package? I can't tell where it's used.

danieladugyan (changes requested):
Neat effect 🤩, just some minor comments

danieladugyan:
Fixed in #107

PR #117, opened 2024-01-02
Add board page and sort committee positions

alfredgrip:
This adds a page for the board, displaying board positions and the person occupying it and some general info. Board positions are displayed in order based on the statutes (stadgan).

Also includes sorting of committee positions for committee pages. Don't know if this is the best way to do it, it relies on hard coded order which is not ideal, but it also doesn't introduce extra fields for each position/committee for the data base.
Highest prio: committee master and vice.
Last prio: general volunteer within that committee.
If position order is not defined, put it last.

TODO: Make the board page info editable by board members. Maybe someone with a bit more experience in this area can take a look at it?


src/lib/utils/committee-ordering/sort.ts
```diff
@@ -36,7 +36,7 @@ export function sortCommitteePos(
   posId2: Position["id"],
   committeeShortName: NonNullable<Committee["shortName"]>,
 ): number {
-  if (!(committeeShortName in committeeEnums)) return 1; // No committee? Assume pos B comes before
+  if (!(committeeShortName in committeeEnums)) return 0; // No committee? Preserve order
```
danieladugyan, line 39: I _think_ this makes more sense, but I'm not sure.

danieladugyan (approved):
I simplified some stuff, but overall I think it looks great. Everything seems to work fine so I think we can we merge.

src/lib/utils/committee-ordering/sort.ts
alfredgrip: Agreed!

alfredgrip:
The reason for my somewhat unintuitive data type was to avoid using `indexOf(...)` to improve sorting speed, but we aren't dealing with big data so I guess it's okay.

danieladugyan:
> The reason for my somewhat unintuitive data type was to avoid using `indexOf(...)` to improve sorting speed, but we aren't dealing with big data so I guess it's okay.

Fair enough – indexOf isn't actually required here either, since TypeScript enums compile to objects:

```js
{
  "1": "dsek.ordf",
  "2": "dsek.vice_ordf",
  "3": "dsek.skattm.mastare",
  "4": "dsek.infu.mastare",
  "5": "dsek.srd.ordf",
  "6": "dsek.cafe.mastare",
  "7": "dsek.nari.mastare",
  "8": "dsek.km.mastare",
  "9": "dsek.aktu.mastare",
  "10": "dsek.sex.mastare",
  "11": "dsek.noll.stab.oph",
  "dsek.ordf": 1,
  "dsek.vice_ordf": 2,
  "dsek.skattm.mastare": 3,
  "dsek.infu.mastare": 4,
  "dsek.srd.ordf": 5,
  "dsek.cafe.mastare": 6,
  "dsek.nari.mastare": 7,
  "dsek.km.mastare": 8,
  "dsek.aktu.mastare": 9,
  "dsek.sex.mastare": 10,
  "dsek.noll.stab.oph": 11
}
```

indexOf is used to make TypeScript happy.

PR #119, opened 2024-01-10
Update keycloak on mandate add and delete

03haax:


src/lib/utils/keycloak.ts
```diff
@@ -0,0 +1,143 @@
...
+  return response[0]!.id;
+}
+
+// turns dsek.sexm.kok.mastare into ['dsek', 'dsek.sexm', 'dsek.sexm.kok', 'dsek.sexm.kok.mastare']
+function getRoleNames(id: string): string[] {
+  const parts = id.split(".");
+  return [...Array(parts.length).keys()].map((i) =>
+    parts.slice(0, i + 1).join("."),
+  );
+}
```
danieladugyan, line 44: IIRC this function is already defined elsewhere (`getDerivedRoles`).

src/lib/utils/keycloak.ts
```diff
@@ -0,0 +1,143 @@
...
+
+async function getGroupId(client: KcAdminClient, positionId: string) {
+  const roleNames = getRoleNames(positionId);
+  const groups = await client.groups.find();
```
danieladugyan, line 48: We should probably cache this, but that can be fixed in a separate issue.

src/lib/utils/keycloak.ts
```diff
@@ -0,0 +1,143 @@
...
+    username: KEYCLOAK_ADMIN_USERNAME || "",
+    password: KEYCLOAK_ADMIN_PASSWORD || "",
+    grantType: "password",
+    clientId: "admin-cli",
+  });
+
+  kcAdminClient.setConfig({ realmName: "dsek" });
+
+  return kcAdminClient;
+}
```
danieladugyan, line 28: We should probably cache this, but that can be solved in another issue.

danieladugyan (approved):

PR #124, opened 2024-01-23
Make database setup script more read-/understandable

fnurkla:


danieladugyan:
Looks good. I fixed some other stuff while we're at it (see commit message).

PR #139, opened 2024-01-31
Remove symbol check and .then(translated) usage

danieladugyan:
This pull request removes the symbol check in the devalue package (used internally by SvelteKit) and consequently allows us to remove all usage of `.then(translated)`.

No more `Cannot stringify POJOs with symbolic keys` 😄


alfredgrip (approved):
LGTM! Nice that you found a solution to this!

PR #142, opened 2024-01-31
138 non conforming folder structures

Isak-Kallini:
All files should now show up regardless of if it's some old folder structure or the new folder structure.


src/routes/[[lang=lang]]/documents/+page.server.ts
```diff
@@ -62,6 +82,25 @@ export const load: PageServerLoad = async ({ locals, url }) => {
...
+    const fileParts = file.id.split("/");
+    const fileName = fileParts[fileParts.length - 1] ?? "unknown";
+    const meeting = fileName!.substring(
+      0,
+      fileName.endsWith("pdf")
+        ? fileName.length - 3
+        : fileName.endsWith("html")
+          ? fileName.length - 4
+          : fileName.length,
+    );
```
Comment (ekvanox) [inline src/routes/[[lang=lang]]/documents/+page.server.ts:98]: I feel like nested ternarys could be confusing here. Also, if we were to handle more file formats it would not be easily expandable for that logic.

Also, is fileName not always defined? If so I don't think we need the non-null assertion ("!").

Maybe something like this? (I haven't tested it)
```
const extensions = ["pdf", "html"]
const meeting = extensions.some((ext) => {
  if (fileName.endsWith(ext)) return fileName.substring(0, fileName.length - ext.length)
}) ?? fileName
```

ekvanox (changes requested):
I saw some minor things I commented on, otherwise LGTM!

src/routes/[[lang=lang]]/documents/+page.server.ts
```diff
@@ -46,7 +57,16 @@ export const load: PageServerLoad = async ({ locals, url }) => {
...
+      filteredFiles = SRDfiles.filter((file) => {
+        const fileParts = file.id.split("/");
+        const meeting = fileParts[fileParts.length - 2] ?? "unknown";
+        return meeting.startsWith("Möte");
+      });
+      oldFormatSRDFiles = SRDfiles.filter((file) => {
+        const fileParts = file.id.split("/");
+        const meeting = fileParts[fileParts.length - 2] ?? "unknown";
+        return !meeting.startsWith("Möte");
+      });
```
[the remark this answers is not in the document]
Comment (Isak-Kallini) [reply src/routes/[[lang=lang]]/documents/+page.server.ts]: Looks much better, fixed!

src/routes/[[lang=lang]]/documents/+page.server.ts
Comment (Isak-Kallini) [reply src/routes/[[lang=lang]]/documents/+page.server.ts]: Good point, just had to modify your code a bit

ekvanox (approved):
Looks good!

PR #153, opened 2024-02-06
114 hover state på like knappen

03haax:


src/routes/[[lang=lang]]/news/LikeButton.svelte
```diff
@@ -8,7 +8,7 @@
   import { isAuthorized } from "$lib/utils/authorization";
 
   export let likers: Member[];
-  $: disabled = isAuthorized(apiNames.NEWS.LIKE, $page.data.user);
+  $: authorized = isAuthorized(apiNames.NEWS.LIKE, $page.data.user);
```
Comment (Isak-Kallini) [inline src/routes/[[lang=lang]]/news/LikeButton.svelte:11]: Not sure if it's just something weird on my end but this doesn't seem to be working, I can like/dislike even when logged out. Looks good otherwise

Comment (03haax) [reply src/routes/[[lang=lang]]/news/LikeButton.svelte:11]: It is because you have the development environment variable set, then you get all permissions.
https://github.com/Dsek-LTH/web/blob/d4b2029e0b74c0c67d958dc289d3b9d757b2d81a/src/lib/utils/authorization.ts#L10

Comment (Isak-Kallini) [reply src/routes/[[lang=lang]]/news/LikeButton.svelte:11]: Gotcha, thanks!

Isak-Kallini (approved):
Gotcha, thanks!

Issue #30, opened 2023-11-10
Notifications


Macludde:
```[tasklist]
### Tasks
- [ ] https://github.com/Dsek-LTH/web/pull/148
- [x] Send notifications, receive them
- [x] Send notifications on specific actions
- [x] Group notifications
- [ ] Push notifications
```

Steindt:
- [ ] Notification bell and dropdown
- [ ] Send notifications, receive them
- [ ] Send notifications on specific actions
- [ ] Group notifications

fnurkla:
@Steindt, are you taking the charge for the push notifications as well?

Steindt:
Yes

PR #169, opened 2024-03-02
Notificationlogic

Steindt:
Backend and updated frontend for web notifications.

Currently only linked to creating news, updating news and pinging as these are the ones implemented so far.

To test for notifications against yourself, uncomment line 95 in _src/lib/utils/notifications/index.ts_.

Part of #30 and completes #168.


src/database/prisma/schema.prisma
```diff
@@ -507,6 +507,9 @@ model Member {
     @@map("members")
 }
 
+/// @@allow('create', has(auth().policies, 'core:member:ping'))
```
Macludde, line 510: Does this mean these rows have to be updated whenever a new way to send notficiations is added? Perhaps we should just set "create" to "_" or whatever "logged in" is?

src/database/prisma/schema.prisma
```diff
@@ -571,6 +574,8 @@ model Payment {
     @@map("payment")
 }
 
+/// @@allow('read', auth().memberId == fromMemberId || auth().memberId == toMemberId)
+/// @@allow('update', has(auth().policies, 'core:member:ping'))
```
Macludde, line 578: Should this include the above check as well, ie that the user is one of the included members in the ping?

src/lib/utils/notifications/index.ts
```diff
@@ -107,50 +118,163 @@ const sendNotification = async (
...
+  //   ) {
+  //     throw error(500, "Failed to send notifications");
+  //   }
+  //   if (pushResult.status === "rejected") {
+  //     throw error(500, "Failed to send push notifications");
+  //   }
+  //   if (databaseResult.status === "rejected") {
+  //     throw error(500, "Failed to save ping to database");
+  //   }
+  // };
```
Macludde, line 158: remove?

Macludde (commented):
I'm still reviewing this, here are some things thus far. When I try to start it locally, I get this error:

```
{
  name: 'PrismaClientKnownRequestError',
  code: 'P2023',
  clientVersion: '5.7.1',
  meta: {
    modelName: 'Notification',
    message: 'Error creating UUID, invalid character: expected an optional prefix of `urn:uuid:` followed by [0-9a-fA-F-], found `n` at 2'
  },
  internalStack: 'PrismaClientKnownRequestError: \n' +
    'Invalid `prisma.notification.findMany()` invocation:\n' +
    '\n' +
    '\n' +
    'Inconsistent column data: Error creating UUID, invalid character: expected an optional prefix of `urn:uuid:` followed by [0-9a-fA-F-], found `n` at 2\n' +
    '    at si.handleRequestError (/Users/ludvigsvedberg/Programming/dsek/web/node_modules/.pnpm/@prisma+client@5.7.1_prisma@5.7.1/node_modules/@prisma/client/runtime/library.js:125:6817)\n' +
    '    at si.handleAndLogRequestError (/Users/ludvigsvedberg/Programming/dsek/web/node_modules/.pnpm/@prisma+client@5.7.1_prisma@5.7.1/node_modules/@prisma/client/runtime/library.js:125:6151)\n' +
    '    at si.request (/Users/ludvigsvedberg/Programming/dsek/web/node_modules/.pnpm/@prisma+client@5.7.1_prisma@5.7.1/node_modules/@prisma/client/runtime/library.js:125:5859)\n' +
    '    at async l (/Users/ludvigsvedberg/Programming/dsek/web/node_modules/.pnpm/@prisma+client@5.7.1_prisma@5.7.1/node_modules/@prisma/client/runtime/library.js:130:10025)'
}
```

Macludde:
I've found some other smaller issues, but I think it would be easier for me to change them locally, push my commits and then you can have a look at them @Steindt

src/lib/utils/notifications/index.ts
```diff
@@ -1,15 +1,15 @@
...
   SUBSCRIPTION_SETTINGS_MAP,
-  type NotificationSettingType,
 } from "$lib/utils/notifications/types";
 import type { Author, PrismaClient } from "@prisma/client";
 import { error } from "@sveltejs/kit";
 
 const DUPLICATE_ALLOWED_TYPES = [
   NotificationType.CREATE_MANDATE,
-  NotificationType.ARTICLE_UPDATE,
+  NotificationType.ARTICLE_REQUEST_UPDATE,
```
Macludde, line 12: I renamed this to be more clear. `ARTICLE_REQUEST_UPDATE` (as it's now called) is meant to be a notification you get when an update has been made to your active article request (not yet built in this version). Such as if it has been approved or rejected.

src/lib/components/Notification.svelte
```diff
@@ -1,32 +1,44 @@
...
-  };
+  export let notification: NotificationItem;
+  export let deleteForm: SuperValidated<NotificationSchema>;
 
-  const handleClick = () => {
-    console.log("Clicking on notification");
-  };
+  const { enhance } = superForm(deleteForm, {
+    id: notification.id.toString(),
+  });
```
Macludde, line 17: We use superForm in this project, and thus we don't use the bultin `enhance` method, feel free to check out how I solved it with this setup for future PRs

src/lib/components/Notification.svelte
```diff
@@ -1,32 +1,44 @@
...
-  export let notificationLink: string;
+  import LiveTimeSince from "$lib/components/LiveTimeSince.svelte";
+  import type { NotificationSchema } from "$lib/zod/schemas";
+  import type { Notification } from "@prisma/client";
+  import type { SuperValidated } from "sveltekit-superforms";
+  import { superForm } from "sveltekit-superforms/client";
+  type NotificationItem = Pick<
+    Notification,
+    "link" | "title" | "message" | "createdAt" | "id"
+  >;
```
Macludde, line 10: Instead of manually declaring the `Notification` object, use the type export from `@prisma/client`. And even better, if you don't use all properties, use `Pick<...>` to select only the properties you actually use. This is nice in the future if we want to do optimization and at a glance see which properties we need to extract from the DB for each page.

src/routes/(app)/read-notifications/+server.ts
```diff
@@ -0,0 +1,45 @@
+import apiNames from "$lib/utils/apiNames";
+import { isAuthorized } from "$lib/utils/authorization";
+import type { RequestHandler } from "./$types";
+
+// Read notification
+const readNotifications: RequestHandler = async ({ locals, url }) => {
```
Macludde, line 6: So I re-built how "read notifications" is done. From an abstract perspective, marking notifications as read is not an explicit user actions, it is done when they click on the bell icon. Previously, this bell icon was within a `<form>` tag, and when clicked, a form would be submitted.

This worked because your browser has javascript enabled, and what superforms (or SvelteKit for that matter) does when javascript is enabled, is that they stop the form submission and instead just sends a fetch request to the server. Regular form requests always lead to a page load. This means notifications would not work for people without JS enabled. Even if those users are few, just semantically this did not make to be a form.

Instead I built a `PUT` endpoint which is called when the bell icon is clicked, marking all notifications as read.

Macludde: Of course, this means non-JS users can not mark their notifications as read. That's a bummer and if necessary, we might want to add some support for that in the future (like a "mark as read" button which is actually a form)

src/routes/+layout.server.ts
```diff
@@ -1,10 +1,28 @@
...
+        orderBy: {
+          createdAt: "desc", // latest first
+        },
+      })
+    : null;
+  return {
+    user,
+    member,
+    notifications: notifications,
+    deleteNotificationForm: await superValidate(notificationSchema),
```
Macludde, line 26: Superform stuff, if you want you can read more on superforms.rocks

src/routes/(app)/NotificationBell.svelte
```diff
@@ -1,32 +1,112 @@
...
+  import type { NotificationSchema } from "$lib/zod/schemas";
+  import type { SuperValidated } from "sveltekit-superforms";
+  import { superForm } from "sveltekit-superforms/client";
+  import { page } from "$app/stores";
+
+  export let notifications: NotificationType[];
+  export let deleteForm: SuperValidated<NotificationSchema>;
+
+  // Get the number of unread notifications, which is then used to indicate the user
+  $: unreadCount = notifications.filter((data) => data.readAt == null).length;
```
Macludde, line 16: This was previously a for loop which mutated an array. This meant reactivity didn't work. In svelte, as opposed to say react, a components script code is only run once, not on every "re-render". Re-renders in svelte work by finding dependencies, for example:
- If you declare a variable with let, it will keep track everywhere that is user
- Variables declared as `$: varName = ...` are called derived and svelte will keep track of which variables they use, the derived property will be updated whenever one of its dependencies changes.

Macludde:
@Steindt I pushed some changes, I also added comments above with the biggest changes.

In summary, the biggest things were:
- We use superForms for form validation, and thus "regular" svelte-kit forms (with the `enhance` imported from svelte) should not be used, instead the `+page.server.ts` should return form data which is used on the client
- `readNotifications` should not be a form action, but instead just an api endpoint as its not an explicit action a user takes. This works in general but breaks for users who do not have JS enabled. A big thing with svelte is the ability for it to work for users who turn off JS (most if not almost all logic is server side). If a user has JS enabled (which are most users) it adds some flair to make the page act more like a SPA (use:enhance for example)
- Import types, use the types from `@prisma/client` instead of declaring the object.

Also, please use camelCase for methods or form actions (the form actions were called `deletenotification` and `readnotification`)

But otherwise, I like what you have done and it worked very well. I'll take a look at pings as well before approving

Macludde:
I would say this is finished now and ready to be merged

Macludde (approved):

src/routes/(app)/read-notifications/+server.ts
Steindt: Good point!

src/lib/utils/notifications/index.ts
```diff
@@ -107,50 +120,52 @@ const sendNotification = async (
...
         title,
         message,
         type,
         link,
         memberId,
-        fromAuthorId: fromAuthor?.id,
+        fromAuthorId: notificationAuthor?.id,
       })),
     }),
   ]);
```
danieladugyan, line 134:
```suggestion
  const databaseResult = await prisma.notification.createMany({
    data: receivingMembers.map(({ id: memberId }) => ({
      title,
      message,
      type,
      link,
      memberId,
      fromAuthorId: notificationAuthor?.id,
    })),
  });
```

src/lib/zod/schemas.ts
```diff
@@ -1,6 +1,9 @@
 import { programmes } from "$lib/utils/programmes";
 import { z } from "zod";
 
+export const emptySchema = z.object({}); // for forms without a body
+export type EmptySchema = typeof emptySchema;
```
danieladugyan, line 5: This is very odd to me. Zod is schema validation library. What's the point of validating an empty object?

src/routes/(app)/members/[studentId]/+page.server.ts
```diff
@@ -60,6 +63,32 @@ export const load: PageServerLoad = async ({ locals, params }) => {
...
+      pingForm: await superValidate(emptySchema),
+      member,
+      publishedArticles: publishedArticlesResult.value ?? [],
+      ping: user
+        ? await prisma.ping.findFirst({
+            where: {
+              OR: [
+                {
+                  fromMemberId: member.id,
+                  toMemberId: user?.memberId,
```
Comment (danieladugyan) [inline src/routes/(app)/members/[studentId]/+page.server.ts:78]:
```suggestion
                  toMemberId: user.memberId,
```

src/routes/(app)/members/[studentId]/+page.server.ts
```diff
@@ -60,6 +63,32 @@ export const load: PageServerLoad = async ({ locals, params }) => {
...
+                },
+                {
+                  fromMemberId: user?.memberId,
```
Comment (danieladugyan) [inline src/routes/(app)/members/[studentId]/+page.server.ts:81]:
```suggestion
                  fromMemberId: user.memberId,
```

src/routes/(app)/members/[studentId]/+page.server.ts
```diff
@@ -60,6 +63,32 @@ export const load: PageServerLoad = async ({ locals, params }) => {
...
+            },
+          })
+        : null,
+    };
+  } catch (e) {
+    throw error(500, "Could not fetch ping");
+  }
 
   const allMemberDoors = await getCurrentDoorPoliciesForMember(
     prisma,
```
Comment (danieladugyan) [inline src/routes/(app)/members/[studentId]/+page.server.ts:94]: Delete unreachable code.

src/routes/(app)/members/[studentId]/pings.ts
```diff
@@ -0,0 +1,152 @@
...
+
+const assertMemberExists = async (
+  prisma: PrismaClient,
+  member: MemberIdentification,
+  errorMsg: string = "Member does not exist",
+) => {
+  try {
+    const foundMember = await prisma.member.findFirst({
+      where: member.memberId
+        ? { id: { equals: member.memberId! } }
```
Comment (danieladugyan) [inline src/routes/(app)/members/[studentId]/pings.ts:75]:
```suggestion
        ? { id: { equals: member.memberId } }
```

src/routes/(app)/members/[studentId]/pings.ts
```diff
@@ -0,0 +1,152 @@
...
+        : {
+            studentId: member.studentId!,
```
Comment (danieladugyan) [inline src/routes/(app)/members/[studentId]/pings.ts:77]:
```suggestion
            studentId: member.studentId,
```

src/routes/+layout.server.ts
danieladugyan, line 1: Notification code shouldn't be in this top-level layout file.

src/routes/(app)/Navbar.svelte
```diff
@@ -7,7 +7,13 @@
   import { routes } from "./routes";
   import { isAuthorized } from "$lib/utils/authorization";
   import { getFullName } from "$lib/utils/client/member";
-  import Notification from "./NotificationBell.svelte";
+  import NotificationBell from "./NotificationBell.svelte";
+  import type { Notification } from "@prisma/client";
+  import type { SuperValidated } from "sveltekit-superforms";
+  import type { NotificationSchema } from "$lib/zod/schemas";
+
+  export let notifications: Notification[] | null;
```
danieladugyan, line 15: Maybe we should get notifications from $page.data.notifications instead? I'm not sure what's best.

src/routes/(app)/+page.server.ts
```diff
@@ -81,3 +86,46 @@ export const load: PageServerLoad = async ({ locals }) => {
     alert: await alert,
   };
 };
+
+// These actions will be accessed from throughout the whole application
+export const actions: Actions = {
+  // Delete single or multiple notifications on database
+  deleteNotification: async ({ locals, request }) => {
```
danieladugyan, line 93: I can't help but feel that noitifications don't really belong here. They don't really have anything to do with this page.

danieladugyan (changes requested):
Looks awesome overall! Sorry for the slow review.

Parts of the code are sometimes a bit hard to understand.  I'm especially concerned with how we want to handle reading and dispatching notifications across the app. Should we use an API-route for reading notifications? Should we use a form action for deleting notifications? It's not obvious which pattern is easier to understand and maintain, but maybe we should decide on one of them.

and I'm not good enough at SvelteKit to know how to solve some of the issues I raised. Page-wide data/actions tend to creep

danieladugyan:
Tagging you @Macludde since I think some of my comments pertain to your changes.

src/routes/(app)/members/[studentId]/+page.server.ts
Comment (Steindt) [reply src/routes/(app)/members/[studentId]/+page.server.ts:94]: Out of place after merge? Moving it to to be included in return

Steindt:
@danieladugyan and I decided to use forms, instead of API calls, when reading notifications to remain consistent in the code.

src/routes/(app)/+page.server.ts
Steindt: Has been moved to separate page

src/routes/(app)/Navbar.svelte
Macludde: I agree

Macludde:
@Steindt we cannot have the bell as a form action. It does not work for users without js enabled at all, because the form action will result in a page reload, which then would close the dropdown.

Macludde:
I changed it so it now uses forms, but also works for non-JS users.

Basically, it no longer automatically reads all notifications when you click the bell icon. At the bottom, there is a "mark all as read button", and if you visit a notification, that notification is marked unread.

src/lib/zod/schemas.ts
Macludde: Superforms require a zod schema, so for a form where no form data is expected, I think using an empty schema makes sense

Macludde:
@danieladugyan would you like to take a quick look and approve if it looks good? I'm somewhat happy with the solution now. It completely uses forms, using a "hidden form" to automatically mark a notification as read. 

I changed the logic, so instead of marking all notifications as read when opening notifications, it now only marks them when read if the user visits the link they are pointing to (for example by clicking on them). I think this makes more sense, and also makes the app a lot easier, because it only has to open the relevant link and the notifications will sort themselves out.

The notification code has been moved to a seperate /notifications page by @Steindt, this moves away some code but does make `/notifications` a "valid" endpoint for the page. Currently, this page is never visited as a JS-enabled user, and for non-JS users it will redirect them back to where they came from.

PR #171, opened 2024-03-02
Class year page

alfredgrip:
This adds a page to view people that started a certain year, and optionally filter per programme.
Since our guild only consists of D, C and VR/AR, when we display "all" we actually just display those.
Also, you need to be signed in to view this page.


danieladugyan (approved):
Awesome! I made some small adjustments, merge if you don't mind them.

PR #173, opened 2024-03-02
Stylistic eslint

alfredgrip:
This PR adds more linting rules and adds a check to the node engine before building


.eslintrc.cjs
```diff
@@ -61,5 +62,7 @@ module.exports = {
         ],
       },
     ],
+    "@typescript-eslint/no-inferrable-types": "off", // We want to be able to be explicit about types, even if they are primitives.
+    "@typescript-eslint/consistent-type-definitions": "off", // Using either type or interface is fine.
```
danieladugyan, line 66: _A train of thought_
1. I've personally maintained a policy of leaving things set to their defaults as far as possible, so this would go against that.
2. On the other hand, perhaps there's no strong arguments for that policy - there's a reason for configuration options.
3. However, stricter rules are nice since they make the codebase more consistent, and turning these rules sets a precedent for turning off rules. After all, we've never turned off any rule and I'm not sure these rules warrant breaking that streak.

I think that train of thought could continue back and forth for a while 😄 TL;DR I think we should remove these lines, but it really doesn't matter very much.

alfredgrip: I agree that we could turn `no-inferrable-types` on again, but I don't know about the other one. Types vs Interfaces is a debate I think we shouldn't have ATM.

danieladugyan: Does it matter? 👀 Personally I've never given it much thought; I only use types, but that's just out of habit.

alfredgrip: I'm team Types as well, but it don't find it necessary to enforce it onto others!

PR #174, opened 2024-03-02
166 show door access on profile

Isak-Kallini:
image

### DIFF
--- a/src/routes/(app)/members/[studentId]/+page.server.ts
+++ b/src/routes/(app)/members/[studentId]/+page.server.ts
@@ -59,12 +59,25 @@ export const load: PageServerLoad = async ({ locals, params }) => {
     throw error(404, "Member not found");
   }
   const member = memberResult.value;
+  const roles = member.doorAccessPolicies.map((d) => d.role).filter(notEmpty);
+  const doorPositions = await prisma.position.findMany({
+    where: {
+      id: {
+        in: roles,
+      },
+    },
+  });
   return {
     form: await superValidate(member, memberSchema),
     member,
+    doorPositions,
     publishedArticles: publishedArticlesResult.value ?? [],
   };
 };
+
+function notEmpty<TValue>(value: TValue | null | undefined): value is TValue {
+  return value !== null && value !== undefined;
+}
 
 const updateSchema = memberSchema.pick({
   firstName: true,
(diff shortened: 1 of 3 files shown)


Isak-Kallini:
@alfredgrip What we talked about last saturday should be fixed now.

src/routes/(app)/members/[studentId]/+page.server.ts
```diff
@@ -58,13 +59,81 @@ export const load: PageServerLoad = async ({ locals, params }) => {
...
+        newDoor.startDate = doorPolicy.startDatetime
+          ? old.startDate === null || doorPolicy.startDatetime > old.startDate
+            ? old.startDate
+            : doorPolicy.startDatetime
+          : null;
+        newDoor.endDate = doorPolicy.endDatetime
+          ? old.endDate === null || doorPolicy.endDatetime < old.endDate
+            ? old.endDate
+            : doorPolicy.endDatetime
+          : null;
```
Comment (alfredgrip) [inline src/routes/(app)/members/[studentId]/+page.server.ts:119]: Not a big fan of nested ternary

src/routes/(app)/members/[studentId]/+page.svelte
```diff
@@ -123,4 +125,35 @@
       {/if}
     </div>
   </div>
+  <div
+    class="col-span-5 sm:col-span-3 sm:col-start-3 md:col-start-1 lg:col-span-2 lg:col-start-1 xl:col-span-1 xl:col-start-1"
+  >
+    {#if isMe}
```
Comment (alfredgrip) [inline src/routes/(app)/members/[studentId]/+page.svelte:131]: Would prefer `{#if isMe && data.allMemberDoors.size > 0}` so we don't render something empty

alfredgrip (changes requested):
Nice work getting all doors, but actually right now we fetch too many 😅 We need to filter mandates based on date so we only account for current mandates. This is what my profile showed without any mandates for this year. (Don't mind the navbar displacement, it only appears during screenshot)

Screenshot 2024-03-04 at 11-56-34 Sebastian Engström D-sektionen

Also, it seems to be inconsistent in the role name displayed next to the door. In the screenshot, both "Hovmästare" and "dsek.sex.hov" is displayed. HOWEVER, the role name is "dsek.sex.hov.mstr" so I guess this isn't an easy fix since "dsek.sex.hov" refers to the subgroup.

Isak-Kallini:
Fixed you comments, and I think I managed to fix groups like "dsek.sex.hov" being displayed.

alfredgrip (changes requested):
Nice job fixing the names as well! Left some minor comments to fix, if those are fixed I think this is good to go. However, the function `getCurrentDoorPoliciesForMember` is long and complex, and I'm a bit scared to touch and refactor it. I don't think this is blocking, but if you wish to try to refactor it that would be nice.

Isak-Kallini:
I tried doing some refactoring, but I'm not sure I can make it much better than it is.

PR #177, opened 2024-03-03
Move keycloak to server only

alfredgrip:
We don't want to accidentally expose Keycloak to the client side. Logic inside a path `$lib/server` will make sure it is run server side only. https://kit.svelte.dev/docs/server-only-modules 
Build step will fail if we import it into client side code, but we don't get any errors from linters unfortunately.


src/lib/server/keycloak/keycloak.ts
danieladugyan, line 1: I would move it to `src/lib/server/keycloak.ts` instead to avoid the `"keycloak/keycloak"` import. `src/lib/server/keycloak/index.ts` is also fine I suppose, although a folder with one file in it seems unnecessary to me.

danieladugyan (approved):
Neat - minio also seems reasonable as a server-only component, but we can fix that later in a new PR.

PR #178, opened 2024-03-03
See committee mandate history

alfredgrip:
Related to #158 

Issues arise for committees where all/some mandates are not calendar year


fnurkla:
This is nice! I'm wondering whether the pagination component would be better positioned below the committee description, as it just affects the mandates and not the markdown above. 

For mandates spanning over multiple years, we could highlight (maybe with italics?) dates that are outside the selected year.

src/lib/components/Pagination.svelte
```diff
@@ -39,6 +39,8 @@ e.g. `?page=1`, `?page=2`, etc. The page number is stored in the URL query.
   export let showFirst = false;
   /** Whether to show the last page button. */
   export let showLast = false;
+  /** Wether to keep the scroll position after a page button has been clicked */
```
fnurkla, line 42: Wether -> Whether

src/routes/(app)/committees/PositionCard.svelte
fnurkla, line 1: This is not bad, but what I meant was that dates in the mandate span not in the selected year should be highlighted. 

For example, if i go to valberedningen and the year 2022, half the people would have mandates that started in 2021 and ended in 2022, where the beginning dates would be highlighted/italic, and the other half would have mandates that started in 2022 and ended in 2023, where the ending dates would be highlighted/italic. 

This way it would be helpful regardless what year you're looking at!

src/routes/(app)/committees/committee.server.ts
```diff
@@ -18,34 +18,53 @@ export type UpdateSchema = typeof updateSchema;
...
  * @param shortName The committee's short name
+ * @param year The year to load the committee for, defaults to current year
  * @returns All data that the every committee load function needs
  */
 export const committeeLoad = async (
   prisma: PrismaClient,
   shortName: string,
+  year = new Date().getFullYear(),
 ) => {
+  // const date = new Date(year, 0, 1);
```
fnurkla, line 29: I think this might be a timezone thing. Be careful that this doesn't break when changing to summer time etc.

fnurkla: If I create a `new Date(2024, 0, 1)` in the JS console in my browser I get jan 1st at 00:00:00, so if It was off by even 1 second it would be shunted into 2023.

src/routes/(app)/committees/PositionCard.svelte
alfredgrip: Just pushed some changes that I think addresses that you mean!

src/routes/(app)/committees/committee.server.ts
alfredgrip: I did some changes now and hopefully they handle this better

Isak-Kallini (approved):
LGTM!

PR #179, opened 2024-03-04
Fix landing page

Macludde:
This PR fixes some annoying bugs on the landing page, more specifically how some items "pop around" on initial load. The problem was that we wrapped the markdown in an anchor tag, and in HTML it is strictly forbidden to have nested anchor tags. The resulting behaviour is therefore undefined and I'm guessing the rendering in the browsers differs from the server or something.

While I was picking around, I fixed/built some other small things. Biggest is that I moved out a lot of code in the landing page into components, and moved all landing page-related files to the new (LandingPage) folder in (app)


src/lib/components/MarkdownBody.svelte
```diff
@@ -3,17 +3,28 @@
...
+      clazz,
+    );
+    if (wrappedLink) {
+      merged = twMerge(
+        "pointer-events-none relative prose-a:pointer-events-auto",
+        merged,
+      );
+    }
+    return merged;
+  })();
```
Isak-Kallini, line 24: Instead of doing this, could we not just avoid using anchor tags on the landing page?
```
const renderer = {
    link(href: string, title: string | null | undefined, text: string) {
      return text;
    },
  };
const customMarked = wrappedLink ? new Marked({ renderer }) : new Marked();
```

and using `customMarked.parse` instead of `marked` on line 30.

I feel like this is both easier to understand and takes less lines of code. This would also solve the problem of being able to tab to links in parts of the text that shouldn't be visible.

Isak-Kallini (commented):
I'm not sure I like the `(LandingPage)` folder. 
- It's not how you're supposed to use groups
- I feel like the natural place to look for the landing page is in the root folder where it was before.

Macludde:
> I'm not sure I like the `(LandingPage)` folder.
> 
> * It's not how you're supposed to use groups
> * I feel like the natural place to look for the landing page is in the root folder where it was before.

I knew this would be a discussion. Personally I feel like this is a good thing to do. The top level folder under `(app)` was getting quite big, with all the components for the layout and now also the components for the landing page. Perhaps a better solution would be to move the components to the $lib folder, but I think that's worse because components in the $lib folder are meant to be re-usable. Components only for a relevant page should be next to the +page.svelte in the file tree

src/lib/components/MarkdownBody.svelte
Isak-Kallini: I don't really think that the links needs to be clickable here. Usually you would click on the article to read the whole thing before clicking any links. I also disagree that all links in the text should be tabbable here, it looks weird if the text scrolls when tabbing, and there is no other way to scroll the text other than tabbing to links.

Isak-Kallini:
> I knew this would be a discussion. Personally I feel like this is a good thing to do. The top level folder under (app) was getting quite big, with all the components for the layout and now also the components for the landing page. Perhaps a better solution would be to move the components to the $lib folder, but I think that's worse because components in the $lib folder are meant to be re-usable. Components only for a relevant page should be next to the +page.svelte in the file tree

What do you think about keeping the landing page stuff in `(app)` and moving all the layout components (Navbar, Footer etc.) to `routes`? That way `(app)` isn't as crowded, the landing page is where you expect it and I feel like it kind of makes sense to have the Navbar components etc. in `routes`.

src/lib/components/MarkdownBody.svelte
Macludde: Fine I can agree with this I guess. I did not understand you meant link not visible, they should of course not be tabbable to. I just meant the links actually visible within the line clamp. But I can settle for having to first go to the article and then press the links.

Macludde:
> LandingPage

Then there's the problem with the `+layout.svelte` file being inside the `(app)` folder but the rest being one step above. That too feels weird. I think perhaps the best solution is moving them out to `$lib` even if it goes against how we work in the rest of the app.

src/routes/(app)/(LandingPage)/Events.svelte
```diff
@@ -0,0 +1,38 @@
...
+          </div>
+        </div>
+
+        <div class="flex min-w-0 flex-col justify-between overflow-hidden">
+          <!-- for screen readers, we icnlude this anchor -->
+          <h2 class="line-clamp-1 text-xl font-bold leading-none">
+            {event.title}
+          </h2>
+          <p class="line-clamp-3 text-ellipsis leading-snug">
+            {markdownToTxt(event.description)}
```
danieladugyan, line 32: This is only safe because the event description has already been sanitized in the `page.server.ts` load function **for this route**. I considered changing this to `markdownToTxt(sanitize(event.description))` to avoid setting a dangerous precedent, but opted not to in the end. Not sure what's the right call, we really need a reliable way of ensuring no XSS vulnerabilities slip through.

danieladugyan:
Great job! The landing page rendering bug was really annoying and it was killing our CLS scores on Vercel's Speed Insights 😅

I made some changes to how the components are named (avoiding doubly qualified names) and reverted/simplified some of the logic around rendering of markdown text excerpts. I think we can merge this now! (but ideally someone should review my changes)

Isak-Kallini (approved):
We discussed the `(LandingPage)` folder in DWWW and decided to move it to (app) and to move the Navbar components etc. to `routes`. 
I also wrapped the article in a div because the margin wasn't clickable in firefox otherwise.

danieladugyan (approved):

PR #180, opened 2024-03-05
Fix profile page layout and door access

Macludde:
#174 was merged but I noticed it overflowed when I tried it on my profile. 

While fixing that I got lost in the logic of how door access policies were fetched and refactored it to be hopefully easier to read, or at least more performant. I hope my changes were fine.

I also re-built how the grid area works on the profile page because the old way (with col-start-X and col-span-Y etc.. was giving me a headache)

### DIFF
--- a/src/lib/utils/member.ts
+++ b/src/lib/utils/member.ts
@@ -1,6 +1,6 @@
 import type { PrismaClient } from "@prisma/client";
-import { getDerivedRoles } from "./authorization";
 import { error } from "@sveltejs/kit";
+import { getDerivedRoles } from "./authorization";
 
 export const getCustomAuthorOptions = async (
   prisma: PrismaClient,
@@ -39,18 +39,30 @@ export const getCustomAuthorOptions = async (
   });
 };
 
+export type MemberDoorPolicies = {
+  name: string;
+  roles: string[];
+  startDate: Date | null;
+  endDate: Date | null;
+}[];
+
 export const getCurrentDoorPoliciesForMember = async (
   prisma: PrismaClient,
-  memberId: string,
+  studentId: string,
 ) => {
-  const [memberResult] = await Promise.allSettled([
-    prisma.member.findUnique({
-      where: {
-        id: memberId,
+  const memberPositionIds = await prisma.position
+    .findMany({
+      select: {
+        id: true,
+        name: true,
+        boardMember: true,
       },
-      include: {
+      where: {
         mandates: {
-          where: {
+          some: {
+            member: {
+              studentId,
+            },
             startDate: {
               lte: new Date(),
             },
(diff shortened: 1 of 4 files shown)


src/lib/utils/member.ts
```diff
@@ -39,117 +39,141 @@ export const getCustomAuthorOptions = async (
...
-      const oldData = allMemberDoors.get(doorPolicy.doorName);
-      const newData = oldData ?? {
-        roles: [...positionNames],
-        startDate: doorPolicy.startDatetime,
-        endDate: doorPolicy.endDatetime,
+        .map((pos) => pos.name);
+      positionsMappedToThisDoor.sort();
+      return {
+        ...policy,
+        roles: positionsMappedToThisDoor || ["Du"],
```
Isak-Kallini, line 172: This never evaluates to `["Du"]`, maybe something like 
`positionsMappedToThisDoor.length > 0 ? positionsMappedToThisDoor : ["Du"]`

alfredgrip (changes requested):
Nice job refactoring the logic for fetching door policies, and breaking out DoorAccess to a component!

However, I'm not a big fan of custom CSS on the `+page.svelte`. IMO, we should stick to using Tailwind to not confuse ourselves. As far as I know, we don't have any custom CSS elsewhere.

src/lib/utils/member.ts
```diff
@@ -39,117 +39,141 @@ export const getCustomAuthorOptions = async (
...
+    [] as MemberDoorPolicies,
+  );
+  const memberDoorPolicies: MemberDoorPolicies = policiesByDoor.map(
+    (policy) => {
+      const positionsMappedToThisDoor = memberPositionIds
+        .filter((pos) =>
+          policy.roles.some(
+            (role) =>
+              pos.id.startsWith(role) ||
+              (pos.boardMember && role === "dsek.styr"),
```
Isak-Kallini, line 165: I don't really understand why "dsek.styr" is in the database in the first place. There's only one entry with that role, for "mauer", and it looks like all positions where "boardMember" is true already have access to mauer. Is there some other purpose for it or could it be removed?

Isak-Kallini (changes requested):
Just some small comments, looks good otherwise!

I agree that the previous layout wasn't very nice to work with, but as Alfred said we don't use custom css anywhere else in the project.

src/lib/utils/member.ts
Macludde: `dsek.styr` is used in more places I think, at least for api access it is used quite a bit. Maybe not as much for doors

Macludde:
So I changed back to only using tailwind, with some componentization I now think it's a bit more clear how the page is structured to new readers

Isak-Kallini (approved):
lgtm

src/lib/utils/member.ts
```diff
@@ -38,6 +38,15 @@ export const getCustomAuthorOptions = async (
     },
   });
 };
+const date = "2022-10-11";
+
+export type MemberDoorPolicies = Array<{
```
Macludde, line 43:
```suggestion

export type MemberDoorPolicies = Array<{
```

src/lib/utils/member.ts
```diff
@@ -65,10 +74,10 @@ export const getCurrentDoorPoliciesForMember = async (
               studentId,
             },
             startDate: {
-              lte: new Date(),
+              lte: new Date(date),
```
Macludde, line 77:
```suggestion
              lte: new Date(),
```

src/lib/utils/member.ts
```diff
@@ -65,10 +74,10 @@ export const getCurrentDoorPoliciesForMember = async (
...
             },
             endDate: {
-              gte: new Date(),
+              gte: new Date(date),
```
Macludde, line 80:
```suggestion
              gte: new Date(),
```

PR #184, opened 2024-03-05
PDF modal

alfredgrip:
Makes the PDFs for governing documents pop up when clicked on.

Not sure about the api path and what specifications it should have, however. Thoughts on this @danieladugyan ?


danieladugyan:
I feel like the download icon should left-aligned, even if the text wraps.

image

danieladugyan (changes requested):
I made some simplifications, but other than that it looks good! There's just a small visual alignment bug left to fix. In the future we should probably use the same PDF modal logic (and the same file component) for all documents, not just governing documents.

danieladugyan:
And as always, feel free to comment my changes if something is odd/wrong.

danieladugyan (approved):

PR #189, opened 2024-03-07
Add verbose door names

fnurkla:
This PR adds a verbose name field to doors to make it easier for people to understand which door access policies and the like refer to.

### DIFF
--- a/src/database/prisma/migrations/20240306193950_add_verbose_door_name/migration.sql
+++ b/src/database/prisma/migrations/20240306193950_add_verbose_door_name/migration.sql
@@ -1,0 +1,18 @@
+-- Add column
+ALTER TABLE "doors" ADD COLUMN "verbose_name" VARCHAR(255);
+
+-- Set values for existing values
+UPDATE "doors" SET verbose_name = 'Border (caféförrådet)'     WHERE "name" = 'border';
+UPDATE "doors" SET verbose_name = 'Fulförrådet'               WHERE "name" = 'ful';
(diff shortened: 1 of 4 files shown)


alfredgrip:
Do we have to allow null values for `verbose_name`? It's not like we insert a new door into our database without having a name for it 🤔

fnurkla:
No, I guess not. I cam change that.

fnurkla:
I realized that I forgot to fix the database migration. Can I just edit it and add in the `NOT NULL` at the end or is that blasphemous?

fnurkla:
Should be good to go now.

Isak-Kallini (changes requested):
I get an error when running `sh dev/setup_db.sh`:
`ERROR: column "verbose_name" of relation "doors" contains null values` from the migration and `The column 'verbose_name' does not exist in the current database.` from doors.ts.
It looks like it tries to add the column but there's already rows in the table and it gives an error because there are null values in the new column. Then it fails in seeding as well since the column wasn't created. 
After a quick google the easiest way to fix this seems like adding a default value to the column. Another option is to initially allow null values, seed it and then alter it to not allow null values, but I'm not sure how to do this with Prisma.

fnurkla:
Alright, I'll check on this tonight!

fnurkla:
Alright, so this should fix that problem. This time I actually checked 😅

Isak-Kallini (approved):
Looks good, nice work!

PR #200, opened 2024-03-12
Clicking the language switcher now always works

danieladugyan:
Previously, there was a seemingly weird behaviour when clicking the language switcher in the top navigation bar. Sometimes you would click it and nothing happened. It turned out to be because the button styling was applied on a div, instead of the actual <a> element that you need to click to switch language.


Isak-Kallini (approved):
LGTM!

PR #206, opened 2024-03-15
Ensure filled board positions are not shown as vacant

danieladugyan:
As can be seen in the "Before" image below, there was a bug where board member positions would be shown as vacant, even though they weren't. It was easily fixed by changing the if-statement that determines if the position is vacant. Aside from fixing the bug, I also tweaked the design of the page.

Before:
dsek vercel app_board

After:
web-git-205-board-positions-are-vacant-dsek vercel app_board

### DIFF
--- a/src/routes/(app)/board/+page.svelte
+++ b/src/routes/(app)/board/+page.svelte
@@ -8,72 +8,81 @@
 
 <PageHeader title="Styrelsen" />
 <!-- TODO: make this editable by board members with Markdown page -->
-<p>
-  Styrelsen ansvarar för den dagliga verksamheten på sektionen. Till styrelsen
-  kan du alltid vända dig om du har frågor, funderingar eller åsikter om
-  sektionen och dess verksamhet. Styrelsen sammanträder på styrelsemöten varje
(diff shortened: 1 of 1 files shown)


src/routes/(app)/board/+page.svelte
```diff
@@ -8,72 +8,81 @@
...
-          </figure>
-          <div class="card-body px-0 text-center">
-            <h2 class="card-title mx-auto">
-              {getFullName(member)}
+<section class="mb-5 space-y-5">
+  <p>
+    Styrelsen ansvarar för den dagliga verksamheten på sektionen. Till styrelsen
+    kan du alltid vända dig om du har frågor, funderingar eller åsikter om
+    sektionen och dess verksamhet. Styrelsen sammanträder på styrelsemöten varje
+    vecka och är öppna för alla medlemmar.
```
fnurkla, line 16: The lats sentence here sounds weird. It's the meetings that are open for all members, not the board it self.

danieladugyan: Agree. It's slightly changed now.

ekvanox (approved):

PR #208, opened 2024-03-18
Move two navbar buttons to drawer on mobile

danieladugyan:
This PR moves two navbar buttons to the drawer on mobile devices to improve the user experience and make better use of the available screen space. Also makes some other changes to the navbar design and some tailwind class name simplifications.

Before (iPhone SE):
dsek vercel app_(iPhone SE)

After (iPhone SE):
localhost_5173_en(iPhone SE)

### DIFF
--- a/src/routes/(app)/+layout.svelte
+++ b/src/routes/(app)/+layout.svelte
@@ -1,11 +1,18 @@
 <script lang="ts">
   import GlobalAlert from "$lib/components/GlobalAlert.svelte";
   import { languageTag } from "$paraglide/runtime";
+  import { onMount } from "svelte";
+  import { themeChange } from "theme-change";
   import "../../app.css";
   import Drawer from "../Drawer.svelte";
   import Footer from "../Footer.svelte";
   import Navbar from "../Navbar.svelte";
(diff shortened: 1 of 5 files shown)


src/routes/(app)/Drawer.svelte
```diff
@@ -12,53 +14,85 @@
...
+          <!-- sun icon -->
+          <span class="swap-off i-mdi-weather-sunny" />
+        </label>
+        <div class="btn btn-ghost btn-sm rounded">
+          {#if languageTag() === "sv"}
+            <a href={i18n.route($page.url.pathname)} hreflang="en">EN</a>
+          {:else if languageTag() === "en"}
+            <a href={i18n.route($page.url.pathname)} hreflang="sv">SV</a>
+          {/if}
+        </div>
```
danieladugyan, line 42: These correspond to the `DarkLightToggle` and `LanguageSwitcher` components. I should probably use those instead, but they don't really work (I had to make some small design changes, i.e making them smaller).

03haax:
Skärmavbild vid 2024-03-26 21-36-53

PR #225, opened 2024-03-21
Add backup list for door access

danieladugyan:
This pull request adds a backup list of student IDs for door access. It also includes testing for the door access endpoint.

### DIFF
--- a/src/routes/(app)/salto/[door]/constants.ts
+++ b/src/routes/(app)/salto/[door]/constants.ts
@@ -1,0 +1,13 @@
+// This needs to be updated at the turn of every year,
+// and it is chosen by the Head of Faculties (Källarmästare).
+// Do not edit it without consulting them.
+export const BACKUP_LIST_OF_STUDENT_IDS = [
+  "em5261ha-s", // Ordförande, Emma Haggren
+  "ph3883ni-s", // KM, Philip Nielsen
+  "es1767st-s", // root, Esbjörn Stenberg
+  "ad2313ad-s", // DWWW-ansvarig, Daniel Adu-Gyan
+  "da6673he-s", // Informationsansvarig, Dag Hemberg
+  "fe0024vi-s", // Skattmästare, Felix Vigren
+  "ol1662le-s", // Revisor, Oliver Levay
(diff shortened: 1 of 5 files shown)


alfredgrip:
I don't know how this is currently implemented on the member-page, but this implementation always give full access to all doors for everyone in the backup-list. As with all access, it could be exploited to do harmful things to our guild and this is why it is important that the guild trusts the people that are granted this access. The positions I'm hesitant to include are "root", "DWW-ansvarig" and "Informationsansvarig". 

root and DWWW-ansvarig are not elected by the guild, but rather the board. However, I understand that these positions are responsible for keeping the actual salto system alive and if it were to break they should be able to fix it, so keeping them on the list is fine.

Informationsansvarig, I guess, is in case we don't have a DWWW-ansvarig, but since we actually have one elected now I don't see why we should include the other.

However, I wonder if an implementation like: "if the endpoint returns an empty list, it should instead return the backup" is better suited? This is not ideal either since if the API is faulty and returns a member which we cannot reach we are in trouble. I don't know which is better, since it's a compromise between the failsafe and the integrity of our access system so I would like this to be discussed more!

src/routes/(app)/salto/[door]/constants.ts
```diff
@@ -0,0 +1,13 @@
...
+// Do not edit it without consulting them.
+export const BACKUP_LIST_OF_STUDENT_IDS = [
+  "em5261ha-s", // Ordförande, Emma Haggren
+  "ph3883ni-s", // KM, Philip Nielsen
+  "es1767st-s", // root, Esbjörn Stenberg
+  "ad2313ad-s", // DWWW-ansvarig, Daniel Adu-Gyan
+  "da6673he-s", // Informationsansvarig, Dag Hemberg
+  "fe0024vi-s", // Skattmästare, Felix Vigren
+  "ol1662le-s", // Revisor, Oliver Levay
+  "al6271gr-s", // Revisor, Alfred Grip
```
Comment (danieladugyan) [inline src/routes/(app)/salto/[door]/constants.ts:12]:
```suggestion
  "ph3883ni-s", // KM, Philip Nielsen
  "es1767st-s", // root, Esbjörn Stenberg
```

PR #231, opened 2024-03-23
Add automatically generated API

danieladugyan:
This PR adds an automatically generated RPC-style API using ZenStack. This allows anyone to read basically any data from our database, provided that they are authorized to do so - the same authorization rules as for our normal database reads apply. I also added OpenAPI documentation that can be used, for instance, to build a Swagger UI to explore our API. 

I have been postponing this PR because there are a lot of things to consider:
- **Do we want to commit to maintaining an API?** Of course it seems trivial right now since it's automatically generated, but what if it breaks? What if we one day want to move away from Prisma and therefore can't use ZenStack to generate an API for us? By opening up an API we're implicitly commiting to maintaining it, at least if people begin building tools based on it.
- **What does support of third-party applications even look like?** Some members may choose to build private tools on this API and that's fine, but others may be build tools used by a wider audience. Should we provide this audience by integrating third-party tools inside the main dsek.se website? I could imagine a "widget" page where users could submit their custom widgets.
In fact, there are probably a few tools right now that could benefit from being integrated on the main dsek.se site (i.e gerda.dsek.se), and by creating an API we are further encouraging this. I think that's awesome! But we don't currently have an infrastructure that supports this. I have been looking at technologies such as [Webpack Module Federation](https://webpack.js.org/concepts/module-federation/) (and its Rollup (Vite) equivalents) that allow runtime integration of third-party code, but we currently have no idea if that's something we want to build upon.
- **Should we provide API documentation?** If we want people to use this then obviously yes. I tried adding interactive API documentation to this PR (using Swagger UI), but I couldn't make it performant enough (the API specification is about 100 000 lines long and there are A LOT of operations; it was either slow or simply crashed).
- **Is RPC och REST more appropriate?** ZenStack allows us to generate two different flavours of API — one that simulates using the Prisma client directly by providing endpoints such as `events/findMany`, and another that is more of a  "standard REST API". I was initially leaning towards REST, but I was swayed towards RPC when I read that ZenStack can generate frontend query hooks for the RPC API. That's obviously a big selling point if we want people to be able to easily fetch data from us in their third-party JS apps. And if we ever want to build a more native React Native app we could use the queries there too.
- **Should automatically generated code be tracked by Git?** It's perhaps only tangential to this PR, but with all these generated files (openapi.yaml) it feels like an annoyance to have to make sure that all commited changes to schema.zmodel also include changes to all of the derivative files (schema.prisma and openapi.yaml are generated from schema.zmodel).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,4 +10,4 @@ vite.config.ts.timestamp-*
 /.vercel
 .zenstack_repl_history
 project.inlang/project_id
-src/database/openapi.yaml
+src/routes/api/openapi.json/openapi.json
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ vite.config.ts.timestamp-*
 /.vercel
 .zenstack_repl_history
 project.inlang/project_id
+src/database/openapi.yaml
--- a/.prettierignore
+++ b/.prettierignore
@@ -8,7 +8,7 @@ node_modules
 !.env.example
 /.vscode
 src/translations/paraglide
-
+src/database/openapi.yaml
 # Ignore files for PNPM, NPM and YARN
 pnpm-lock.yaml
 package-lock.json
--- a/.prettierignore
+++ b/.prettierignore
@@ -8,7 +8,7 @@ node_modules
 !.env.example
 /.vscode
 src/translations/paraglide
-src/database/openapi.yaml
+src/routes/api/openapi.json/openapi.json
 # Ignore files for PNPM, NPM and YARN
 pnpm-lock.yaml
 package-lock.json
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "vite-node": "^1.1.0",
     "vitest": "^1.2.2",
     "vitest-mock-extended": "^1.3.1",
-    "zenstack": "^1.7.1",
+    "zenstack": "^1.11.1",
     "zod": "^3.22.4"
   },
   "type": "module",
@@ -88,7 +88,8 @@
     "@tiptap/pm": "^2.1.13",
     "@tiptap/starter-kit": "^2.1.13",
     "@vercel/speed-insights": "^1.0.10",
-    "@zenstackhq/runtime": "^1.7.1",
+    "@zenstackhq/runtime": "^1.11.1",
+    "@zenstackhq/server": "^1.11.1",
     "croppie": "^2.6.5",
     "dayjs": "^1.11.10",
     "expo-server-sdk": "^3.7.0",
--- a/package.json
+++ b/package.json
@@ -108,7 +108,8 @@
     "tailwind-merge": "^2.2.0",
     "theme-change": "^2.5.0",
     "tiptap-markdown": "^0.8.8",
-    "uuid": "^9.0.1"
+    "uuid": "^9.0.1",
+    "vite-plugin-cjs-interop": "^2.1.0"
   },
   "pnpm": {
     "patchedDependencies": {
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "@types/uuid": "^9.0.8",
     "@typescript-eslint/eslint-plugin": "^6.15.0",
     "@typescript-eslint/parser": "^6.15.0",
+    "@zenstackhq/openapi": "^1.11.1",
     "autoprefixer": "^10.4.16",
     "concurrently": "^8.2.2",
     "daisyui": "^4.4.24",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -47,8 +47,11 @@ dependencies:
     specifier: ^1.0.10
     version: 1.0.10(@sveltejs/kit@2.4.3)(svelte@4.2.8)
   '@zenstackhq/runtime':
-    specifier: ^1.7.1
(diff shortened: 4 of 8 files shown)


danieladugyan:
Requesting review from @Dsek-LTH/dwww since I think this requires a broader discussion. I'd also be interested to hear your thoughts @fnurkla since you were interested in this feature.

Isak-Kallini:
First, I don't understand how to use the api. For example I tried 
`https://web-git-230-auto-generated-rest-api-dsek.vercel.app/api/model/events/findfirst` 
but only get 
`{"message":"An unhandled error occurred: TypeError: RPCApiHandler is not a function"}`
for all queries I try. If i try it locally I get 
`{"message":"An unhandled error occurred: TypeError: __vite_ssr_import_10__.default is not a function"}` 
instead.

- **Do we want to commit to maintaining an API?** I really like the idea of having an api. But I have no idea how much work it would be to maintain. If we someday would want to move away from Prisma it would definitely make it harder, but not impossible I think. We would probably have to maintain two versions of the api or something for a while, so users would have time to switch.
- **What does support of third-party applications even look like?** I feel like this would be pretty low priority for now, but the idea with third party widgets sounds really cool.
- **Should we provide API documentation?** Yes I think so, it also seems like there are a lot of different tools for generating documentation for openapi so we should be able to find something that works for us.
- **Is RPC och REST more appropriate?** I think RPC looks really nice.
- **Should automatically generated code be tracked by Git?** I don't know what's best practice, but spontaneously I feel like generated files shouldn't be tracked.

fnurkla:
> Requesting review from @Dsek-LTH/dwww since I think this requires a broader discussion. I'd also be interested to hear your thoughts @fnurkla since you were interested in this feature
Thank you!

I think that this feature is of importance because it enables guild members to create their own applications for other members. In this use case, I think the most important parts are to be able to _authenticate and authorize_ members (to be able to couple data to users in other applications[^1]) and to be able to _read data_ from our database (if authorized to do so [^2]).

>     * **Do we want to commit to maintaining an API?** [...]
This is a good question. I think that we want to do so at some point for the reasons I just mentioned. How much do we expect the structure of our data and consequently the API to change? Or at least change in a non backward-compatible way? In semantic versioning terms, do we expect major version bumps all the time? Or just minor versions?

Maybe this is something that has to wait until after the launch? Or say that launch also is the 1.0.0 version of the API and go from there?

>     * **What does support of third-party applications even look like?** [...]

I don't think considerations about this should block adding an API. It's an interesting discussion, but I think that we wait and see what people do first, and then adapt accordingly.

>     * **Should we provide API documentation?** [...]

I think this ties very much into the next question. I assume that the REST option is less extensive?

>     * **Is RPC or REST more appropriate?** [...]

From my point of view, the REST option is nice because it is approachable. If all students learn about REST APIs in the PVG course, having that option would make it easy for people to get up and running. This doesn't necessarily mean that the RPC option is bad, but I think that it makes things a lot harder to approach, especially if we have >100 000 different endpoints without documentation :sweat_smile:.

I also think that developing the website in this repo is generally a pretty good outlet for web development for members, but I think other forms of development do not have the same outlet right now and that's why I think this issue is important. In that way I think prioritizing web development for this isn't very important.

>     * **Should automatically generated code be tracked by Git?** [...]
Regardless of how we do this, if we want an API we need to version control it in some way, otherwise it will be impossible to develop for, at least in the medium to long term. I am partial to semantic versioning as it is very clear about what different version bumps imply. How we choose to version our API might have implications for whether we track these files or not.

[^1]: Take gerda for example, being able to couple drafts to a user could be enabled by being able to "log in" through the guild website.
[^2]: This could for example be used to make a CLI news reader for the guild.

alfredgrip:
I think this is interestring but I would just like to speak for myself as the creator of Gerda. I understand the point that if we could authenticate users, a database could be set up for drafts and prior work on Gerda or even share documents between users. However, in this case I myself haven't got any plans to integrate this even if we were to have this API. My goal is to keep Gerda simple and easy to maintain since some day in the future I probably won't be able to maintain it any longer and it should therefore be easy to get into that project. I wouldn't strictly object if someone else wishes to integrate this into Gerda, but I don't think it's necessary.

Other thoughts:
An API could definitely be a useful tool for enabling other users to create own applications, but I'm a little hesitant towards this automatically generated one (partly because I don't fully understand the implications of it). Could another alternative be to manually create endpoints we want? Like authentication and read news/events? If we then notice people actually using it, we could consider expanding the API even further.

danieladugyan:
> First, I don't understand how to use the api. For example I tried `https://web-git-230-auto-generated-rest-api-dsek.vercel.app/api/model/events/findfirst` but only get `{"message":"An unhandled error occurred: TypeError: RPCApiHandler is not a function"}` for all queries I try. If i try it locally I get `{"message":"An unhandled error occurred: TypeError: __vite_ssr_import_10__.default is not a function"}` instead.

@Isak-Kallini This should be fixed now, thanks to help from one of the Zenstack maintainers. Try https://web-git-230-auto-generated-rest-api-dsek.vercel.app/api/model/event/findFirst for instance.

src/database/openapi.yaml
fnurkla, line 1: Should we actually track this 100 000+ line generated file?

PR #234, opened 2024-03-24
Translate landing page

danieladugyan:
This pull request implements translations for the landing page. Note that I had to use `@html` to render rich translated text (i.e bold parts in the translated text). I think this will become easier to implement once the paraglide.js library completes [#913](https://github.com/opral/monorepo/discussions/913).


03haax:
Looks good!

PR #238, opened 2024-03-25
Various minor alert fixes

### LINKED ISSUE #193: Deleting a global alert gives error 500
If you delete a global alert at the moment, you get served a 500 page. The alert is deleted as it should, but it still feels confusing for the user.

### LINKED ISSUE #237: Alerts should show up on all pages

danieladugyan:
Closes #193, closes #237.

This PR fixes some small issues with global alerts:
- They are now visible on all pages, not just the landing page.
- The date formatting is now correctly localized by using the dayjs library (timezone was wrong).
- Users now get a "Are you sure?" dialog before deleting an alert.
- Users no longer get an `500 Internal Server Error` when deleting alerts.
- The message text in the table of alert now only shows the message in the current language.
As a sidenote, our current way of handling translations doesn't really support showing the Swedish and the English version on the same page. Instead, it would show the two languages correctly when viewing the page in Swedish, but in English it would simply show the English text in both columns.

### DIFF
--- a/src/lib/utils/softDelete.ts
+++ b/src/lib/utils/softDelete.ts
@@ -1,0 +1,21 @@
+import { Prisma } from "@prisma/client";
+import { error } from "@sveltejs/kit";
+
+/**
+ * Wrap a soft delete function in a try-catching block.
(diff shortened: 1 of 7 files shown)


Steindt (approved):
Looks good!

PR #242, opened 2024-03-26
219 kunna se folks dsek mailadresser p deras profiler

03haax:


src/lib/server/keycloak/index.ts
```diff
@@ -185,6 +185,13 @@ async function hasEmail(email: string) {
   return user.filter((u) => u.email === email).length > 0;
 }
 
+async function getEmail(username: string) {
+  if (!enabled) return false;
```
ekvanox, line 189: Should this not return null? Or is there a reason for a false here?

In other places we just use:
```
if (!enabled) return;
```

ekvanox (changes requested):
Looks good, just have some questions :)

ekvanox (approved):

PR #251, opened 2024-04-09
247 webshop/part1: Add database model for tickets system

danieladugyan:


src/database/schema.zmodel
```diff
@@ -797,3 +804,178 @@ model LastKeycloakUpdate {
...
+  description String
+  descriptionEn String?
+  type ShoppableType
+
+  authorId String @db.Uuid @deny("update", true) // never allow updating wh created the shoppable
+  author Member @relation(fields: [authorId], references: [id], onDelete: NoAction, onUpdate: NoAction)
+
+  price Int
+  availableFrom DateTime @map("available_from") @db.Timestamptz(6) @default(now())
+  availableTo DateTime? @map("available_to")
```
fnurkla, line 830: Why no `@db.Timestamptz(6)` here? Am I missing something?

src/database/schema.zmodel
```diff
@@ -797,3 +804,178 @@ model LastKeycloakUpdate {
...
+  titleEn String?
+  description String
+  descriptionEn String?
+  type ShoppableType
+
+  authorId String @db.Uuid @deny("update", true) // never allow updating wh created the shoppable
+  author Member @relation(fields: [authorId], references: [id], onDelete: NoAction, onUpdate: NoAction)
+
+  price Int
+  availableFrom DateTime @map("available_from") @db.Timestamptz(6) @default(now())
```
fnurkla, line 829: Maybe consider having this nullable. This would allow for authors to create drafts for items before they know when they are to be published.

Maybe that could be changed later, though.

src/database/schema.zmodel
```diff
@@ -797,3 +804,178 @@ model LastKeycloakUpdate {
...
+  // Unreadable if removed, unless author or manager
+  @@deny("read", shoppable.removedAt != null && !(auth().memberId == shoppable.authorId || has(auth().policies, "webshop:manage")))
+  // author (and managers) can update and delete questions
+  @@allow("update", auth().memberId == shoppable.authorId || has(auth().policies, "webshop:manage"))
+  @@allow("delete", auth().memberId == shoppable.authorId || has(auth().policies, "webshop:manage"))
+
+  @@map("item_question")
+}
+
+model ItemQuestionAnswer {
```
fnurkla, line 899: I think we should choose a different name for this if it reflects the different options to choose between, I immediately got it mixed up with ItemQuestionResponse. `ItemQuestionVariety`or `ItemQuestionAlternative` is a lot better imo.

src/database/schema.zmodel
fnurkla: Also, items that are never meant to be sold? Like café tickets would be consumables, but they need a shoppable as well, but it should never be published!

src/database/schema.zmodel
```diff
@@ -797,3 +804,178 @@ model LastKeycloakUpdate {
...
+  shoppable Shoppable @relation(fields: [shoppableId], references: [id], onDelete: NoAction, onUpdate: NoAction)
+
+  memberId String? @db.Uuid 
+  member Member? @relation(fields: [memberId], references: [id])
+  externalCustomerEmail String? // if purchased by a non-member
+  externalCustomerCode String? // if purchased by a non-member, unique code used in link as security
+
+  createdAt DateTime @default(now()) @map("created_at") @db.Timestamptz(6)
+  purchasedAt DateTime? @map("purchased_at") @db.Timestamptz(6)
+  // customer or webshop:consume can always consume from unconsumed state. manager can "unconsume" from consumed state
```
fnurkla, line 959: We talked about not having the user be able to consume consumables on their own, to prevent people from accidentally spending tickets before an event, for example. Instead it's people at the event who spend them for you. What do you think of that?

fnurkla (commented):
Some things I thought about while reading through the changes.

src/database/schema.zmodel
Macludde: Good catch. Honestly, the only reason we have `@db.Timestamptz(6)` everywhere is due to it being imported from the old database. I don't really think it is, or isn't, necessary. But I guess we could be consequential.

src/database/schema.zmodel
Macludde: I think this is a different issue which could be handled in a future PR. Perhaps together with "ItemAccessPolicies" where we could se custom policies for who can buy a Shoppable. Café tickets could then have the policy "webhsop:item:not_visibile" or something.

I don't want to increase the scope of these PRs unecessarily

src/database/schema.zmodel
Macludde: Yeah I agree, naming is hard... What about `ItemQuestionOption`?

Macludde:
@fnurkla your comments have been resolved

src/database/schema.zmodel
fnurkla: That sounds good as well.

src/database/schema.zmodel
fnurkla: Yes exactly, but then its not the owner who consumes it, right? It would be the host who does that

Macludde: sorry my comment might be weird. Anyways, I removed it so you can't consume it on your own.

What I meant is that I thought of having something like that, but I was also thinking that we might want to add an override feature. But that could be a feature in the future.

Macludde:
@fnurkla @danieladugyan can this be merged?

fnurkla (approved):

danieladugyan:
> @fnurkla @danieladugyan can this be merged?

I have a pending review as well.

Macludde (approved):

PR #257, opened 2024-04-10
247 build purchase system

Macludde:
This PR (which might get split up as well) integrates payment into the webshop. It integrates with stripe (you need some keys to try it out live).

This is how it works:
- Go to cart, show user the cart items (frontend not built)
- When they press purchase:
  - Get cart items, verify attributes (still available, still stock left, not expired, not already purchased etc)
  - Handle/fix eventual errors from previous checks)
  - Create a Stripe Payment Intent (basically a request for a one-time charge). Which payment methods that are available is set in the stripe dashboard.
  - Run a check and verify (within a transaction) that none of the cart items have during the function call been part of another purchase, and then save the intent id in the database.
  - Get the client secret and send it to the frontend
- The frontend uses this clientSecret and our stripe public key to show the stripe-built payment element frontend. Basically the form for inputting card details, or paying with swish/klarna etc.
- When they press purchase, and the payment goes through, a webhook call is sent to us from stripe. If the payment succeeded, we use the "intent id" saved in the database to mark the correct cart items as purchased. If it failed, we don't.

Of course, if the cart only includes free items, they are purchased immidiately. A transaction fee is also added to all purchases. Sadly, some payment methods have higher fees than others, and it is illegal in the EU to charge different amounts as a merchant depending on the users' choice of payment method. I chose to use the fee for Swish, as I think it will be the most common payment methods.

# Tests
There are quite a few tests for this, of course. I think I've covered a lot of cases and edge cases, but certainly not all of them, only the ones I (and ChatGPT) could think of. I don't think we'll ever build a 100% secure and fail-safe payment system, but I think its okay that once in a blue moon an admin will have to go in and adjust somethings manually via stripe.

With that said, if you do come up with new edge cases feel free to say so, or make them new issues to be tackled in the future. A lot of the edge cases I've covered is already very rare to happen (like a user having two browsers open and clicking "purchase" on both at the same time).

## Known issue
I'm not an expert at vitest, I think I've come up with a somewhat nice way of building these tests. I'm mocking stripe and using live prisma. This is done by at each test generating new data, with a unique id. There is also a unique member for each test suite. The problem with using live prisma is that we are vulnerable to side effects in tests. While I've tried to ensure we don't, there are a few tests which "sometimes fail, at random".

My best guess is that they fail depending on the order of other tests being run, or if another test is being run at the same time. Vitest makes sure to run all tests in a file sequentially (if we don't use the `concurrent` modifier), but files are run in parallel for performance. I have not been able to figure out which tests are making other tests fail, but I've seen that it is most common (or only) when in the "anonymous user context".


Macludde:
Okay I think I fixed the issue with side effects in tests, or at least this specific one. While writing my documentation above I realized I use the same `externalCode` for anonymous users in different test files. Giving them each a unique external code fixed it. In production, of course all users will have a unique external code (or at least "v4 uuid unique")

PR #261, opened 2024-04-16
Improve door access query

Isak-Kallini:
Since #180 seems to not be worked on atm (I think I remember @Macludde saying he wanted to redo the whole page), I think we can at least merge the improved door access query.

### DIFF
--- a/src/lib/utils/member.ts
+++ b/src/lib/utils/member.ts
@@ -39,12 +39,12 @@ export const getCustomAuthorOptions = async (
   });
 };
 
-export type MemberDoorPolicies = {
+export type MemberDoorPolicies = Array<{
   name: string;
   roles: string[];
   startDate: Date | null;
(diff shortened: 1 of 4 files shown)


src/lib/utils/member.ts
```diff
@@ -39,117 +39,149 @@ export const getCustomAuthorOptions = async (
...
 
+export type MemberDoorPolicies = Array<{
+  name: string;
+  verboseName: string | undefined;
+  roles: string[];
+  startDate: Date | null;
+  endDate: Date | null;
+}>;
+
 export const getCurrentDoorPoliciesForMember = async (
```
alfredgrip, line 50: Perhaps this function could use some tests, but it's not blocking IMO

alfredgrip (approved):
LGTM! Left a comment but feel free to merge

PR #279, opened 2024-04-20
30 push notifications

Steindt:
Added push notifications


src/lib/utils/notifications/index.ts
```diff
@@ -32,6 +32,9 @@ type SendNotificationProps = BaseSendNotificationProps &
       }
   );
 
+// Need permissions to read expo tokens and send notifications without sharing expo tokens to the public
+const prisma = new PrismaClient();
+
```
Macludde, line 37: Should use the `authorizedPrismaClient` which has now been introduced in a file with the same name. Makes it more clear when reading the code that it is a "dangerous" client, and that it uses the same instance everywhere.

src/lib/utils/notifications/index.ts
```diff
@@ -136,9 +136,39 @@ const sendNotification = async (
...
+  if (expoTokens != undefined) {
+    sendPushNotifications(
+      expoTokens.map((token) => token.expoToken),
+      title,
+      message,
+      type,
+      link,
+    );
+  }
```
Macludde, line 171: should be extracted to a new function. I think the same goes for the `databaseResult` query above, so basically we would have two async functions:
- `sendInAppNotifications(memberIds[], title...)`
- `sendPushNotifications(memberIds[], title,...)`

These could then be called and awaited simulatenously, reducing the time of this function. I think that's crucial because it's going to be run often and sometimes having to send a lot of notifications at once (like some news tags). I think that's how it was done before and is why it's called `databaseResult` (because it awaited a `Promise.allSettled([...])` call

src/lib/utils/notifications/push.ts
```diff
@@ -1,5 +1,5 @@
 import { error } from "@sveltejs/kit";
-import Expo, { type ExpoPushMessage } from "expo-server-sdk";
+import { Expo, type ExpoPushMessage } from "expo-server-sdk";
 import { dev } from "$app/environment";
 
 async function sendPushNotifications(
```
Macludde, line 5: I know this is not from this PR, but I REALLY think we should enable push notification security and use our expo private key here. As it is right now, given someones expo push notification ANYONE could send a push notification to anyone of our users.

Macludde (changes requested):
Just looked through this. Looks good, have some ideas for refactoring the now very large `sendNotification` method. Also, as mentioned in the relevant comment, I think we should enable push notification security and think this PR is a good place to do it. Just adding an .ENV variable basically

src/lib/utils/notifications/push.ts
Steindt: Sounds like a good idea. Unfortunately don't know how to do this, would be great if you could do a commit with the required changes.

fnurkla:
We should fix push notification security ASAP, but it shouldn't be blocking. Merging.

PR #280, opened 2024-04-21
Add simple ticket page

danieladugyan:


fnurkla (approved):
Looks fine

src/routes/(app)/shop/tickets/[slug]/+page@(app).svelte
Comment (fnurkla) [inline src/routes/(app)/shop/tickets/[slug]/+page@(app).svelte:1]: Wait, why is this screen wide instead of containered like basically all other pages on desktop?

fnurkla:
Yep, fine now!

PR #281, opened 2024-04-22
Recurring events

Isak-Kallini:
Currently only implemented creating and deleting recurring events, I'll create an issue for other features


src/routes/(app)/events/Event.svelte
```diff
@@ -1,8 +1,13 @@
 <script lang="ts">
   import DateSpan from "./DateSpan.svelte";
   import MarkdownBody from "$lib/components/MarkdownBody.svelte";
-  import type { Event } from "@prisma/client";
-  export let event: Event;
+  export let event: {
```
Steindt, line 4: You can use Pick<Type, parameters> and still use the type from @prisma/client

```
export let event: Pick<
    Event,
    | "title"
    | "startDatetime"
    | "endDatetime"
    | "shortDescription"
    | "description"
  >;
```

Isak-Kallini: Good catch, thanks!

Steindt (approved):
Apologize for late response. Nice work, seem to work as expected, LGTM!

src/database/schema.zmodel
```diff
@@ -349,9 +349,47 @@ model EventComment {
...
+  alarmActive Boolean? @default(false) @map("alarm_active")
+  tags Tag[] @relation("recurring_event_tags")
+  events Event[]
+  author Member @relation(fields: [authorId], references: [id], onDelete: NoAction, onUpdate: NoAction, map: "events_author_id_foreign")
+
+  @@allow("create", has(auth().policies, "event:create"))
+  @@allow("read", has(auth().policies, "event:read"))
+  @@allow("update", has(auth().policies, "event:update"))
+  @@allow("delete", has(auth().policies, "event:delete"))
+}
```
Macludde, line 386: Just looking through this PR real quick and I don't think we should have this kind of duplication, at least I think a lot of these fields are just copy-pasted from the Event model. Instead I think this should inherit from events, adding some extra fields necessary for repeated events.

Isak-Kallini: Yeah you're probably right. My idea with this originally was that recurring events would read from the event in the RecurringEvent table, so that updating for example the title for all events would just be changing this one event. I didn't end up doing it this way though so yes I should remove a lot of the fields.

PR #301, opened 2024-04-23
Fix misc. access policy usage

Isak-Kallini:
Update policies for the admin page so everyone doesn't have read access, and so that the admin button in the navbar isn't visible to everyone.
I also added access checks to the email-alias page so you can't see the form if you can't create/edit aliases.

I also found that I couldn't delete/read my notifications, this was because it used LOGGED_IN (="_") from apiNames, which is used in the db to say that all logged in users has some access policy, but it's not an access policy itself so `authorize(apiNames.LOGGED_IN, user)` would always throw an error.

### DIFF
--- a/src/lib/utils/apiNames.ts
+++ b/src/lib/utils/apiNames.ts
@@ -7,6 +7,7 @@ const crud = <prefix extends string>(base: prefix) =>
   }) as const;
 
 const apiNames = {
+  ALERT: "alert",
   NEWS: {
     ...crud("news:article"),
     MANAGE: "news:article:manage",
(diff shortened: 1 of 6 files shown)


Steindt:
I'm able to see the admin button when not logged in. Can also create alerts, delete them, edit email alias when not logged in. 
Screenshot from 2024-04-25 11-31-40

Isak-Kallini:
> I'm able to see the admin button when not logged in. Can also create alerts, delete them, edit email alias when not logged in. 

Did you comment out `return getAllAccessPolicies(prisma);` on line 18 in hooks.server.helper.ts and `if (dev && user) return true;` on line 10 in authorization.ts? Also, looking at the vercel deployment it seems to be working..

Steindt (approved):
Apologize for late response. Yeah, I missed to comment those out, LGTM!

Macludde:
@Isak-Kallini is this PR ready for deployment? I would like to have this deployed tomorrow for our ticket sale

Isak-Kallini:
> @Isak-Kallini is this PR ready for deployment? I would like to have this deployed tomorrow for our ticket sale

Yes it should be, merging now

PR #307, opened 2024-04-29
Add ticket admin page with sold tickets info

Macludde:


src/lib/server/shop/addToCart/reservations.ts
```diff
@@ -229,6 +229,7 @@ const moveReservationsToCart = async (
   reservationsToMove: ConsumableReservation[],
   updateOrder = true,
 ) => {
+  if (reservationsToMove.length === 0) return;
```
fnurkla, line 232: Do you ever encounter this case in the wild? Or should it error in some kind of way so that we know that it happened?

src/lib/server/shop/payments/stripeMethods.ts
```diff
@@ -135,3 +139,27 @@ export const ensurePaymentIntentState = async (
...
+  stripeIntentId: string,
+  amount: number,
+) => {
+  try {
+    const intent = await getPaymentIntent(stripeIntentId, {
+      expand: ["latest_charge"],
+    });
+    if (intent.status !== "succeeded") return; // refund can be seen as not necessary
+    // if already refunded, or disputed and lost
+    if ((intent.latest_charge as Stripe.Charge | null)?.refunded) return; // already refunded
```
fnurkla, line 153: Maybe throw some kind of error here so that we know when were trying to refund something already refunded?

src/lib/server/shop/payments/purchase.ts
```diff
@@ -224,16 +226,22 @@ const purchaseCart = async (
...
+            },
+            data: {
+              stripeIntentId: intent.id,
+              priceAtPurchase: consumable.shoppable.price,
+            },
+          }),
+        ),
+      );
+      if (results.some((result) => result.status === "rejected")) {
+        throw new Error("Det gick inte att spara transaktions id och pris");
```
fnurkla, line 243:
```suggestion
        throw new Error("Det gick inte att spara transaktions-ID och pris");
```

src/routes/(app)/shop/tickets/[slug]/manage/download-csv/+server.ts
```diff
@@ -0,0 +1,95 @@
...
+    minimumFractionDigits: 0,
+    maximumFractionDigits: 2,
+    currency: "SEK",
+    currencyDisplay: "code",
+  });
+  for (const consumable of consumables) {
+    const member = consumable.member;
+    const name = member
+      ? `${member.firstName} ${member.lastName}`
+      : "Anonym användare";
```
Comment (fnurkla) [inline src/routes/(app)/shop/tickets/[slug]/manage/download-csv/+server.ts:77]: Does this mean we don't get the names of anonymous users when they buy tickets?

fnurkla (commented):
A few small things

src/lib/server/shop/addToCart/reservations.ts
Macludde: Yes this is quite common to be true. If `moveQueueToCart` or `moveQueueForwardOneStep` is called when there is no active queue this if statement will be true. So we stop it here really just for performance reasons, avoiding unecessary calls to the database which doesn't do anything.

src/lib/server/shop/payments/stripeMethods.ts
Macludde: I don't agree. Usually, this would be the case if one of two things happened:
1. The admin pressed the refund button before, but for some reason the server crashed AFTER refunding the stripe payment but BEFORE removing the consumable from our database.
2. Someone refunded the user on the stripe dashboard, and THEN went to our dsek admin page to refund the user.

In both cases, I think just pretending like the refund went through is preferable.

src/routes/(app)/shop/tickets/[slug]/manage/download-csv/+server.ts
Comment (Macludde) [reply src/routes/(app)/shop/tickets/[slug]/manage/download-csv/+server.ts:77]: Exactly, because we never ask them. However, there is a plan to ask them for their email. And you could always add "ItemQuestions" with questions like "What is your name?", "Do you have a food preference?" etc... As of now it has not been implemented but I've also temporarily disabled anonymous purchases in a separate PR.

fnurkla (approved):

PR #309, opened 2024-04-29
Temporarily disable anonymous purchases

Macludde:
Well, this PR is not really "temporary", nor does it "disable access to purchase for anonymous users". What it does, however, is make sure all pages and components respect the api access policy "webshop:purchase" (which they didn't before).

So to actually temporarily disable purchases for anonymous users:
- Merge this amazing PR
- Remove "*" from "webshop:purchase"
- Add "_" to "webshop:purchase"


danieladugyan (approved):
LGTM based on a quick glance

PR #310, opened 2024-05-01
Translate tickets and shop system

Macludde:
While doing this I found another issue, which is how redirect and goto works with paraglide.

Paraglide automatically preprocesses all <a> tags to include the language tag at the start of the href="", but it does not do that for `goto(...)` or `redirect(...)` calls. For that you should use i18n.resolveRoute(...) like this: `goto(i18n.resolveRoute("/shop/cart"));` See paraglide docs: [https://inlang.com/m/dxnzrydw/paraglide-sveltekit-i18n#customizing-link-translation](https://inlang.com/m/dxnzrydw/paraglide-sveltekit-i18n#customizing-link-translation)

I chose to create a `goto` and `redirect` wrapper to be used throughout the project instead, which automatically calls i18n.resolveRoute.


danieladugyan (approved):
Good job! Didn't have time for more than a quick glance but LGTM!

src/lib/components/SearchBar.svelte
```diff
@@ -1,5 +1,5 @@
 <script lang="ts">
-  import { goto } from "$app/navigation";
```
danieladugyan, line 2: At some point we should add eslint warnings for these "untranslated" imports (goto and redirect). There are already similar custom ESLint rules in place so it should be quite easy.

Macludde: Good idea!

PR #318, opened 2024-05-02
Use existing policies for DoorAccessPolicies

fnurkla:


src/database/schema.zmodel
```diff
@@ -298,9 +298,10 @@ model DoorAccessPolicy {
   endDatetime DateTime? @map("end_datetime") @db.Timestamptz(6)
   door Door @relation(fields: [doorName], references: [name], onDelete: NoAction, onUpdate: NoAction, map: "door_access_policies_door_name_foreign")
 
-  @@allow("create", has(auth().policies, "core:access:door:update"))
-  @@allow("read", true)
```
danieladugyan, line 302: Could this break the `/salto/<door>` API? [It reads doorAccessPolicies](https://github.com/Dsek-LTH/web/blob/4e06e05175e5e8ea2f909303fbbe2b7a4f6b3c88/src/routes/(app)/salto/%5Bdoor%5D/%2Bserver.ts#L7).

fnurkla: After some cursory testing, it "breaks" only when the `*` access policy for `core:access:door:read` is removed. And when it breaks, it gives `["ph3883ni-s","es1767st-s"]`.

danieladugyan: Alright no worries then!

PR #323, opened 2024-05-05
Continuous deployment

Isak-Kallini:
The deploy pipeline is now working, but I'd like someone with more experience to look over what I've done so I haven't done anything crazy.

The pipeline starts when you create a new release, the github action builds and zips what's needed (as far as I can tell, I did notice what was already on the dsek-web server didn't include some of what I've included so would appreciate input here). It then adds the build.zip to the release and triggers jenkins. I added a secret in github that's needed in the url to trigger jenkins so that not anyone can trigger jenkins.

Jenkins then fetches the build.zip from github and copies it to web-beta. I added the plugin ssh-agent to jenkins to handle ssh credentials. Then ssh to web-beta running the deploy-script, which removes everything in /server/dsek-web except node_modules, build.zip and .env. It then unzips the build.zip, runs npm i and restarts the service.

For all this to work I created the user transferUser on web-beta and added ssh-credentials (I copied this from dsekweb, the current website). I added the user to a group called deploy and gave this group rwx access to /server/dsek-web. I also added the lines 
```
## Allow users in deploy group to restart dsek-web.service
%deploy ALL=NOPASSWD: /bin/systemctl restart dsek-web.service
```
to /etc/sudoers to give the deploy group access to restart dsek-web.service.

I also changed dsek-web.service to run `node /server/dsek-web/build` instead of `node /server/dsek-web`.

I also copied what was in /server/dsek-web into /server/temp-backup/dsek-web before I started if we need to revert to that.

### DIFF
--- a/deploy-script.sh
+++ b/deploy-script.sh
@@ -6,6 +6,6 @@ shopt -s extglob
 rm -r !(node_modules|build.zip) 
 
 unzip build.zip -d .
-npm i --ignore-scripts
+pnpm i --ignore-scripts
 
 sudo systemctl restart dsek-web.service 
--- a/deploy-script.sh
+++ b/deploy-script.sh
@@ -1,0 +1,11 @@
+#! /bin/bash
+cd /server/dsek-web 
+
+# Remove everything but the node_modules folder, .env is also excluded by default
+shopt -s extglob
+rm -r !(node_modules|build.zip) 
+
+unzip build.zip -d .
+npm i --ignore-scripts
+
+sudo systemctl restart dsek-web.service 
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -22,7 +22,7 @@ jobs:
 
       - uses: montudor/action-zip@v1
         with:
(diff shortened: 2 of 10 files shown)


deploy-script.sh
```diff
@@ -0,0 +1,11 @@
+#! /bin/bash
+cd /server/dsek-web 
+
+# Remove everything but the node_modules folder, .env is also excluded by default
+shopt -s extglob
+rm -r !(node_modules|build.zip) 
+
+unzip build.zip -d .
+npm i --ignore-scripts
```
stagrim, line 9: Om man kör npm istället för pnpm så ignoreras väl lock-filen?

Isak-Kallini:
Fixed usage of env variables in latest commit.

If you don't use import from `$env/dynamic` the environment variables will be hardcoded into the build files, so env variables which are secret and different in production need to be imported dynamically. 

For this to work it also needs to be started with `node -r dotenv/config build` (in the dsek-web.service file on the server).

https://kit.svelte.dev/docs/modules#$env-dynamic-private
https://kit.svelte.dev/docs/adapter-node#environment-variables

danieladugyan (approved):

Issue #313, opened 2024-05-02
The price label has too low contrast in light mode


erik-stein:
![Screenshot from 2024-05-02 13-08-01](https://github.com/Dsek-LTH/web/assets/61153756/092cb120-c98c-4e24-bab1-e5d19019f534)

Macludde:
#325 closes this

I, however, still think it should have some sort of color different from grey/black/white

PR #325, opened 2024-05-06
Misc cosmetic improvements

### LINKED ISSUE #312: The text when hovering on ticket cards is unreadable in light mode
image

fnurkla:
Fixes #312 and alleviates #313.

### DIFF
--- a/src/lib/components/Price.svelte
+++ b/src/lib/components/Price.svelte
@@ -13,6 +13,6 @@
   });
 </script>
 
-<span class={twMerge("font-semibold text-success", clazz)}>
+<span class={twMerge("font-semibold", clazz)}>
   {formatter.format(price / 100)}
 </span>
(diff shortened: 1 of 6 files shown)


src/lib/components/Price.svelte
```diff
@@ -13,6 +13,6 @@
   });
 </script>
 
-<span class={twMerge("font-semibold text-success", clazz)}>
+<span class={twMerge("font-semibold", clazz)}>
```
Macludde, line 16: I do not agree with this design decision. I agree it doesn't look great in light mode, but I think we should find another greenish color for light mode.

Macludde: @fnurkla Can we get the nice green color back please? I think this design looks too "drowned out" and boring. 
Screenshot 2024-05-10 at 22 06 28

PR #333, opened 2024-05-08
Fix relative path in SearchBar goto

stagrim:
### Fixed

SearchBar falsely redirects to /. Fix to redirect to current page and update the URLSearchParams

### Steps to reproduce Bug

1. Go to [https://new.dsek.se/en/news](https://new.dsek.se/en/news)
2. Type something in the search bar
3. Magically get transported to / or /home


BlixtSamme (approved):
LGTMOG🐲

PR #338, opened 2024-05-12
Translate a bunch of stuff

fnurkla:
This PR translates a bunch of stuff. It avoids going into translating big chunks like news, events or members, but otherwise covers a lot of bases. I've tried to keep the translation names short but specific.


Steindt (approved):
LGTM

PR #339, opened 2024-05-13
Add different light mode colors

fnurkla:
Add other colors to light mode and add back the green text to prices in the webshop.


Isak-Kallini (approved):
Looks much better than before! Yellow is a bit dark, but ig it's necessary for contrast with white.

PR #344, opened 2024-05-14
239 search bar

03haax:


Steindt (changes requested):
Looks good! Works nicely, just some small stuff. Search text isn't really visible with light mode and the whole search container, names in the search results, becomes a bit blurry compared to the rest of the website. Additionally, the search bar doesn't work without javascript.
Screenshot from 2024-05-14 19-28-58
Screenshot from 2024-05-14 19-48-18

fnurkla:
Works nicely now. I would like it to be possible to just [click away to close the modal](https://daisyui.com/components/modal/#dialog-modal-closes-when-clicked-outside), though.

Isak-Kallini:
Really nice! Small suggestion: maybe nicer to use [Daisy UI modal](https://daisyui.com/components/modal/#dialog-modal-closes-when-clicked-outside)?

03haax:
> Really nice! Small suggestion: maybe nicer to use [Daisy UI modal](https://daisyui.com/components/modal/#dialog-modal-closes-when-clicked-outside)?

I found the modal to have some unwanted behavior such as briefly flashing at the top of the page and hiding the navbar when scrolled down  :upside_down_face:  :upside_down_face:  However changed to the Dialog element so its more like standard HTML

Isak-Kallini (approved):
Too bad the Daisy UI modal didn't work. In that case, looks good!

PR #353, opened 2024-05-15
Modifiy lightmode color scheme

Isak-Kallini:
I got a flash of inspiration when visiting [inet.se](https://www.inet.se/) and accidentally turned on light mode, they have a darker background and light cards, the opposite of what we have, which I think looked pretty good. We all know our light mode looks bad, so I did some experimenting and came up with this which I think looks way better than before. 
Input is very welcome though since it's a very subjective matter.

old:
image

new:
image


fnurkla (approved):
Cool! Looks a lot better imo!

PR #355, opened 2024-05-15
Fixed dark mode flash when reloading

### LINKED ISSUE #349: Configure dark/light mode properly
If you are in light mode everything flashes dark when reloading. This probably has to do with ssr. 
Another problem is that you can't use `dark:` in tailwind on first load, it seems to depend on localstorage.

Steindt:
Adds data-theme attribute from localstorage before rest is loaded, should also fix :dark.

Fixes and closes #349


fnurkla (approved):
Seems to work well! LGTM

PR #356, opened 2024-05-15
Settings and edit bio

Steindt:
Fixes some issues with overlapping elements in the settings page on smaller screens, fixed access for settings page and edit-bio.

Screenshot of overlapping elements:
Screenshot from 2024-05-15 12-14-41


src/routes/(app)/settings/SubscriptionTags.svelte
```diff
@@ -12,7 +13,11 @@
     .filter((t) => t != "");
 </script>
 
-<div class="flex flex-col items-center sm:grid sm:grid-cols-2">
+<div
+  class={languageTag() == "sv"
+    ? "flex flex-col items-center md:grid md:grid-cols-2"
+    : "flex flex-col items-center xl:grid xl:grid-cols-2"}
```
fnurkla, line 19: Is this really the best way to do this? Feels kinda sketchy

Steindt: It's pretty sketchy, the english names are usually longer than the swedish ones but it either looks pretty empty or cramped if you only have 1 of them. Could otherwise decide on a longer settings page by not having notifications beside tags, but there will be a bit more scrolling if that's ok.

fnurkla: Fair enough. It looks fine when I look at it, so it's not outright wrong.

alfredgrip:
Just pushed some changes, thought it was easier than doing a bunch of code reviewing. Let me know if you agree with my changes or not!

src/routes/(app)/members/[studentId]/edit-bio/+page.server.ts
```diff
@@ -1,13 +1,12 @@
...
 import { error, fail } from "@sveltejs/kit";
 import { message, superValidate } from "sveltekit-superforms/server";
 import type { Actions, PageServerLoad } from "./$types";
-import { authorize } from "$lib/utils/authorization";
+import { redirect } from "$lib/utils/redirect";
 
 export const load: PageServerLoad = async ({ locals, params }) => {
   const { prisma, user } = locals;
-  authorize(apiNames.MEMBER.UPDATE, user);
+  if (!user || !user.memberId) throw redirect(302, "/home");
```
Comment (Macludde) [inline src/routes/(app)/members/[studentId]/edit-bio/+page.server.ts:9]: Why? I think it's weird to redirect someone to home screen instead of just showing error. It can make the page seem confusing

Comment (Macludde) [reply src/routes/(app)/members/[studentId]/edit-bio/+page.server.ts:9]: You've also removed the verification that a user can actually edit this person's bio.

Comment (Macludde) [reply src/routes/(app)/members/[studentId]/edit-bio/+page.server.ts:9]: But I see, the old system seems to not check if params.studentId === user.studentId, in which case you should be able to edit. Otherwise you cannot edit your own bio.

src/routes/(app)/settings/+page.server.ts
```diff
@@ -1,36 +1,30 @@
...
-        },
-      })
-    : [];
+  if (user?.memberId == null) throw redirect(302, "/home");
 
+  const subscriptionSettings = await prisma.subscriptionSetting.findMany({
+    where: {
+      memberId: user.memberId,
+    },
+  });
```
Macludde, line 13: If a user is not logged in, instead of redirecting them we should show an error to be clearer. This redirect doesn't make much sense. `throw error(401, "Not logged in")` instead (can't remember exact syntax, check other `page.server.ts` files or even the `authorize` function

src/routes/(app)/settings/+page.server.ts
```diff
@@ -44,46 +38,56 @@ export const actions: Actions = {
     const { user, prisma } = locals;
     const form = await request.formData();
     if (!user) return fail(401, { form });
```
Macludde, line 40: this should probably use zod and superforms to handle type validation, like we do in other form inputs

src/routes/(app)/settings/+page.server.ts
```diff
@@ -44,46 +38,56 @@ export const actions: Actions = {
...
+    }
+    const subscription: string[] = [];
+    for (const v of form.getAll("subscription")) {
+      if (typeof v !== "string") return fail(400, { form });
+      subscription.push(v);
+    }
+    const tags: string[] = [];
+    for (const v of form.getAll("tag")) {
+      if (typeof v !== "string") return fail(400, { form });
+      tags.push(v);
```
Macludde, line 59: could be replaced with three one-liners like this:
`const push: string[] = form.getAll("push").filter((v) => typeof v === "string");`

If we want the form to fail for invalid types, I think zod should help with that. Then we don't have to do the `hack` above and could just do `const pus = form.data.push`;

src/routes/(app)/settings/+page.server.ts
```diff
@@ -44,46 +38,56 @@ export const actions: Actions = {
...
     }
+
+    const memberId = user.memberId;
+    if (memberId == null) return fail(400, { form });
```
Macludde, line 63:
```suggestion
    if (memberId == null) return fail(401, { form });
```

Macludde: 400 is Bad Request, ie something wrong with the form data. 401 is Unauthorized, ie not logged in.

src/routes/(app)/settings/+page.server.ts
```diff
@@ -44,46 +38,56 @@ export const actions: Actions = {
...
+      data: {
+        subscribedTags: {
+          set: tags.map((tag) => {
+            return {
+              id: tag,
+            };
+          }),
+        },
+      },
+    });
```
Macludde, line 91: Should probably be done in a transaction in-case one of these fails. Because in that case a user might lose all their subscription settings but none would be added or something. I know this isn't written in this PR but always good to improve our code.

Also, we don't have to remove all of the user's settings, only remove the ones not in the form data.

src/routes/(app)/settings/+page.svelte
```diff
@@ -96,7 +96,7 @@
       <SubscriptionTags {tags} subscribedTags={subscribedTags.subscribedTags} />
     </div>
     <button
-      class="btn absolute bottom-0 mb-4 mt-4 w-full max-w-xl bg-primary"
+      class="btn absolute bottom-0 mb-4 mt-4 w-full max-w-xl bg-primary text-white"
```
Macludde, line 99: Across the page we have white text color for buttons with primary background. I do, however, think we should btn-primary instead of bg-primary to have consequential styling across the page.

src/routes/(app)/settings/SubscriptionTags.svelte
Macludde: I think we should find some solution with auto columns, min-content or something like that. We never know what the tags will have as names in the future.

Macludde (changes requested):
Some nice things but I think a lot of this code was quite outdated before this PR and some steps can be taken to make it even better.

danieladugyan:
This has been open for a long while so I'm closing it as stale. Feel free to re-open and merge when issues/conflicts are resolved.

Steindt:
Fixed some of the issues brought up, such as refactoring to use transaction when deleting and creating the subscriptions. 

Have made it so the text inside the containers wrap, this way we don't have to worry too much about texts becoming too long  in the future. Not the prettiest but can change the design later once we find something better.

Fixed spacing for push notifications, should look the same for both languages. And the text should wrap if it for some reason it too long on smaller screens.

PR #371, opened 2024-05-24
Change light mode primary to one with correct chroma

fnurkla:
In this PR I change the light mode primary color to one that is one of the correct chroma.

I did this by plugging `#F280A1` into https://oklch.com/ and then reducing the lightness until we reached WCAG AAA for all text elements on the landing page.


Isak-Kallini (approved):
Looks much better!

PR #379, opened 2024-06-08
Translate events

fnurkla:
This translates basically all event routes. It does not, how ever, translate notifications.


src/routes/(app)/events/InterestedGoingList.svelte
```diff
@@ -2,33 +2,60 @@
...
+                    name1: getFullName(going[0]!),
+                    name2: getFullName(going[1]!),
+                    others: going.length - 2,
+                  })
+                : m.events_interestedGoing_two({
+                    name1: getFullName(going[0]!),
+                    name2: getFullName(going[1]!),
+                  }),
+          })
+      : "";
```
Macludde, line 29: this feels a bit cluttered. I don't know for paraglide but for other i18n libs there is usually native support for pluralisation. I'm not sure if that works here, but nonetheless this should be moved to perhaps a function or something with clearer flow. It took me a few seconds to understand what happens here, and if we just split it up to a few lines of codes with some if-elses I think if will be much clearer.

src/routes/(app)/events/InterestedGoingList.svelte
```diff
@@ -2,33 +2,60 @@
...
+                ? m.events_interestedGoing_threeOrMore({
+                    name1: getFullName(interested[0]!),
+                    name2: getFullName(interested[1]!),
+                    others: interested.length - 2,
+                  })
+                : m.events_interestedGoing_two({
+                    name1: getFullName(interested[0]!),
+                    name2: getFullName(interested[1]!),
+                  }),
+          })
```
Macludde, line 49: same as previous, very cluttered.

src/routes/(app)/events/[slug]/+page.svelte
```diff
@@ -24,7 +26,9 @@
 </script>
 
 <svelte:head>
-  <title>{event.title} | D-sektionen</title>
+  <title
+    >{languageTag() === "en" && event.titleEn ? event.titleEn : event.title} | D-sektionen</title
```
Comment (Macludde) [inline src/routes/(app)/events/[slug]/+page.svelte:30]: I think this is done automatically way down in the abstraction layers so to say. Basically, depending on the language, "event.title" either represents the data stored in "title" OR "titleEn". I think it's done in one of the hooks in the root folder.

Macludde (changes requested):
LGTM, one small refactoring I would like is for a very big nested ternary expression to be split up into a function or code block to clearer. Also there is a code line which does basically `lang == "en" ? titleEn : title` (pseudocode), which is unecessary because that is done for all `xxxxx` vs `xxxxxEn` fields in prisma, it's a custom prisma extension we have.

After those are fixed this can be merged.

Macludde:
Nice, much cleaner. LGTM!

Macludde (approved):

PR #384, opened 2024-06-10
Translate news

fnurkla:


src/routes/(app)/news/LikersList.svelte
```diff
@@ -2,23 +2,33 @@
...
+          ? m.news_threeOrMore({
+              name1: getFullName(likers[0]!),
+              name2: getFullName(likers[1]!),
+              others: likers.length - 2,
+            })
+          : m.news_two({
+              name1: getFullName(likers[0]!),
+              name2: getFullName(likers[1]!),
+            })
+      : "";
```
Macludde, line 23: See my comment on #379 for a similar thing, nested ternaries are bad and this gets quite complex and hard to read.

Macludde (changes requested):
LGTM, one small thing which is the same as a comment in #379 .

Macludde (approved):

PR #399, opened 2024-07-02
Make new dsek.se work with updated dsek app

Macludde:
When we switched website the app (obviously) broke, mostly because it points to app.dsek.se (which this is not hosted on) and even after switching to dsek.se there are a lot of special featues which are required for the app to be usable.

This PR solves that. It starts off by looking at incoming requests' User-Agent header, which for the app is `DSEK-APP`, the new `isApp` local variable is then true, which is used to render stuff differently. This is, for example, a new app header bar, a bottom navigation bar, a special home screen and account screen and some other small stuff.

This PR also handles notifications, so the native app can communicate with the website to upload it's notification token. The native app can also detect if the page is running in dark or light mode, to change the status bar (clock, battery and cellular connection at the top of a phone) color accordingly. Finally, the app sends info about it's "safe zones". For example, on newer iOS devices (since about 6-7 years back) there is a notch at the top and a bar at the bottom of the screen. Content can be shown underneath this, but usually have to take them into account as to not hide content. This is done by again giving the size of the notch and bar in the user agent, so the header and bottom navigation can be padded accordingly.

All in all it shouldn't affect the regular site. If you want to try out the app-version you can set the MOCK_IS_APP=true environment variable in .env.local


src/app.html
```diff
@@ -1,5 +1,5 @@
 <!doctype html>
-<html lang="%paraglide.lang%">
+<html lang="%paraglide.lang%" data-theme="">
```
danieladugyan, line 2: Instead of replacing the string `data-theme=""`, consider replacing something like `%theme%` instead. Although it may be unlikely, `data-theme=""` could be used in other instances (as a somewhat related example, it's used for hard-coding the theme for nollning pages).

```suggestion
<html lang="%paraglide.lang%" data-theme="%theme%">
```

src/hooks.server.ts
```diff
@@ -174,11 +175,49 @@ const apiHandle = zenstack.SvelteKitHandler({
...
+const themeHandle: Handle = async ({ event, resolve }) => {
+  const theme = event.cookies.get("theme");
+
+  if (!theme || !themes.includes(theme)) {
+    return await resolve(event);
+  }
+
+  return await resolve(event, {
+    transformPageChunk: ({ html }) => {
+      return html.replace('data-theme=""', `data-theme="${theme}"`);
```
danieladugyan, line 209: See comment in `src/app.html`

src/lib/components/nav/PageHeader.svelte
```diff
@@ -0,0 +1,19 @@
...
+
+  export let title: string | undefined;
+  let clazz = "";
+  export { clazz as class };
+</script>
+
+<SetPageTitle {title} />
+
+<!-- If it's an app, the page title is shown at the top, in the header -->
+{#if !$page.data.isApp}
```
danieladugyan, line 14: The comment says `"If it's an app"`, but the condition says `if not page.data.isApp`.

src/routes/Navbar.svelte
```diff
@@ -106,9 +99,9 @@
           {shopItemCounts}
         />
       {:else}
-        <button class="btn btn-ghost" on:click={() => signIn("keycloak")}>
+        <LoadingButton class="btn btn-ghost" onClick={() => signIn("keycloak")}>
           {m.navbar_logIn()}
-        </button>
+        </LoadingButton>
```
danieladugyan, line 104: Why is this a loading button? Does it make a difference when all `signIn` does is redirect the user?

src/routes/UserMenu.svelte
```diff
@@ -71,10 +70,13 @@
...
-      <button class="btn btn-ghost justify-start" on:click={() => signOut()}>
-        <span class="i-mdi-logout size-6 text-primary" />
-        {m.navbar_userMenu_logOut()}</button
+      <LoadingButton
+        class="btn btn-ghost justify-start"
+        onClick={() => signOut()}
       >
+        <span class="i-mdi-logout size-6 text-primary" />
+        {m.navbar_userMenu_logOut()}
+      </LoadingButton>
```
danieladugyan, line 79: Same question as for `signIn`

danieladugyan (commented):
Overall it looks good to me, I haven't looked too much into the details. I just have some small comment based on a cursory glance.

src/lib/components/nav/PageHeader.svelte
Macludde: Well, the comment doesn't describe the code below, but instead a comment to why this logic is there. I think the logic speaks for itself. "If not an app, show large h1 text with title". But why? That is described in the comment

src/routes/Navbar.svelte
Macludde: Not really, it actually does a fetch first. At least I noticed this quite clearly when I was on a slow network. I pressed the sign in button, it took about 2-3 seconds of nothing, then it navigated to the login page.

src/routes/UserMenu.svelte
Macludde: Same answer

src/routes/Navbar.svelte
danieladugyan: Huh, I didn't know that. [You're right](https://github.com/nextauthjs/next-auth/blob/main/packages/frameworks-sveltekit/src/lib/client.ts#L42)!

src/hooks.server.ts
danieladugyan, line 209:
```suggestion
      return html.replace("%theme%", theme);
```

src/lib/components/nav/PageHeader.svelte
danieladugyan: Aha of course. I'll clarify it slightly.
```suggestion
<!-- If it's an app, the page title is shown at the top instead, in the header -->
{#if !$page.data.isApp}
```

danieladugyan (approved):

PR #415, opened 2024-08-01
Add questions on webshop consumables

Macludde:
When creating a ticket the author can add "ItemQuestions"s. A question has a title, a description and a type. Right now they can be "free text" and "multiple choice". For multiple choice, a number of options can be created. An option can have an "extraPrice" which will act as an extra charge on the final price.

Example:
Ticket for a sitting
Question: Drinks package?
Type: multiple choice
Options:
- Yes, (extra price: +90 SEK)
- No

When a user adds the ticket to their cart (or after the lottery) they will be asked the questions. The price will update if they press yes. They can view and edit their answers if they want, up until they purchase the ticket.

Once purchased, they can see their answers on the inventory page for the item. Admins see their answers in the manage view and when exporting as CSV.


danieladugyan (approved):
I can't really review PRs this large, but feel free to merge.

Macludde:
> I can't really review PRs this large, but feel free to merge.

Fair enough. But it works, and if it doesn't, we'll fix it later.

PR #416, opened 2024-08-01
Fix profile pic upload

Macludde:
Haven't really figured out why you're not able to upload pictures currently


Macludde:
i figured it out:
We were saving profile pictures in /{studentId}/profile-picture but we were looking at public/{studentId} when showing user's their options. fixed

PR #469, opened 2024-08-25
461 redirect nollanu or dseksenolla to dseksenollning

BlixtSamme:
Added the logic in the hook.server.ts because similiar logic for redirecting to onboarding existed in the file, i do not know if there are any downsides to this or if it should be implemented better another way.

Here is the logic:

- /nollning redirects to /nolla until 2024-09-06 23:00
- /nolla redirects to /nollning after 2024-09-06 23:00


src/hooks.server.ts
```diff
@@ -159,6 +159,19 @@ const databaseHandle: Handle = async ({ event, resolve }) => {
...
+    Date.now() < Date.parse("2024-09-06T23:00:00")
+  ) {
+    redirect(302, "/nolla");
+  }
+  if (
+    i18n.route(event.url.pathname) == "/nolla" &&
+    Date.now() > Date.parse("2024-09-06T23:00:00")
+  ) {
+    redirect(302, "/nollning");
+  }
```
Macludde, line 173:
```suggestion
  if (
    i18n.route(event.url.pathname) == "/nolla" &&
    Date.now() > Date.parse("2024-08-26T00:00:00")
  ) {
    redirect(302, "/nollning");
  }
```

1. I see no reason to redirect /nollning to /nolla as nolla.nu goes to /nolla
2. The date is supposed to be the first monday of the nollning (/nollning has a pre-reveal mode which still has support for ticket sales and messages)

Macludde (changes requested):

Macludde:
I moved the logic to a /nolla/+layout.server.ts as there is no reason to do it in hooks. The reason we do it fo onboarding is because it requires of the current user etc.

I also changed the date to midnight tonight.

PR #502, opened 2024-10-01
Only show Interested/Going buttons if user is logged in

### LINKED ISSUE #499: Possible to check Interested and Going on Events when logged out
Users that aren't signed in can still see and press the buttons for "Interested" and "Going" on events. Doing this when logged out causes an "Internal error".

fgren:
Fixes #499


Isak-Kallini (approved):
LGTM!

Issue #282, opened 2024-04-22
Recurring event features


Isak-Kallini:
I've implemented creating and deleting recurring events in #281, but some more feature are still needed:

- [ ] Edit recurring events
- [ ] Edit/delete this and all following events
- [ ] Handling detached events somehow

Macludde:
@Isak-Kallini is this finished? It seems like the support is there but I'm not sure

Isak-Kallini:
Last i checked it wasn't, but I don't know if anyone has worked on this. To clarify, edit recurring events means editing *all* events at once, which wasn't possible last I checked. It's possible to delete all events in a series, but not to only deleting events from a certain point in time, which I think could be useful. With handling detached events (ie singular events that have been moved or something), iirc a detached event is just ignored from the rest of the recurring events as it is right now, this could probably be handled in a nicer way.

But yeah, if this is implemented it should be finished. I haven't checked since I made the original recurring events

PR #503, opened 2024-10-02
Edit recurring events

Isak-Kallini:
part of #282
This pr implements editing recurring events where you can edit all events or a selected event and all in the future. It also adds the option to delete a selected event and all in the future.


src/lib/events/server/actions.ts
```diff
@@ -185,26 +198,49 @@ export const updateEvent: Action<{ slug: string }> = async (event) => {
...
+  } else if (editType === "FUTURE") {
+    await prisma.event.updateMany({
+      where: {
+        recurringParentId: existingEvent.recurringParentId,
+        startDatetime: { gte: existingEvent.startDatetime },
+      },
+      data: {
+        ...eventData,
+        authorId: undefined,
+      },
```
Isak-Kallini, line 231: It doesn't seem possible to update relations with updatemany (https://github.com/prisma/prisma/issues/3143). We could update the tags but that would be a lot of db calls. 
Suggestions welcome!

Macludde (approved):
LGTM!

PR #505, opened 2024-10-09
Booking page quality of life update

oliverlevay:
The "Create booking" button is hard to see, this is fixed by updating the color to our primary color.

When you change the start date you expect the end date to adjust to also be after the start date. This is fixed with this update.

It is hard to see which box belongs to which bookable, so a border is added to differentiate.


danieladugyan (approved):
Awesome, thanks! ⭐

PR #509, opened 2024-10-09
Added edit button users own news/event articles

### LINKED ISSUE #493: No edit button on own news article
Commit 5f988f813b8c5677ac202d85ab8f50909e8bb107 made it so everyone can edit their own news articles. The condition for when the edit button is shown on an article was not changed however, so there is no way for someone to edit thier own article (unless they go to [slug]/edit manually, which they probably don't know exists unless they have read the code).

In the code it looks like it applies to events too but I could not check that as I do not have an event that I created.

Fiery-132:
Fixed #493


Isak-Kallini (approved):
Lgtm, good job!

Issue #485, opened 2024-09-09
Link shortener tag filtering doesn't work


erik-stein:
When choosing a tag for filtering the links list, I get to an error page: 422 Expected array, received string.
The url for the resulting page is https://www.dsek.se/admin/links?search=&tags=valleb23

PR #513, opened 2024-10-10
Suggested fix for Link Shortener's tag filtering

Fiery-132:
Fix for #485. Converts `tags` parameters into an `Array`, as is expected by the `paramsSchema`.


danieladugyan:
LGTM!

I did push a minor change though, let me know if you don't agree with it!

danieladugyan (approved):

Issue #525, opened 2024-10-22
Can't add phadders to phaddergroups


danieladugyan:
I.e can't add mentors to phaddergroups from 2020 https://www.dsek.se/committees/nollu?year=2020

PR #529, opened 2024-10-22
Fixed issue 525

RufusKoggRojder:
Fixed "Can't add phadders to phaddergroups" issue #525 by adding year as an optional search parameter in entity search.


danieladugyan (approved):
Awesome ✨

PR #535, opened 2024-10-24
Refactor "has active mandate" to reduce db stress

alfredgrip:
Benchmarks show improvements by 5% 😎

While it's technically faster, it's not too much which is depressing since I thought this would be substantially faster. I guess the overhead from Prisma is what actually is costly, while the query is basically negligible. But still, I think this is good practice

### DIFF
--- a/src/lib/server/loadHomeData.ts
+++ b/src/lib/server/loadHomeData.ts
@@ -18,7 +18,8 @@ export const loadHomeData = async ({
 
   /* files - subject to change */
 
-  const year = new Date().getFullYear();
+  const now = new Date();
+  const year = now.getFullYear();
   const files = await fileHandler.getInBucket(
     user,
     PUBLIC_BUCKETS_DOCUMENTS,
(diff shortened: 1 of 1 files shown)


src/lib/server/loadHomeData.ts
```diff
@@ -110,15 +111,36 @@ export const loadHomeData = async ({
...
+    prisma.mandate.findFirst({
+      where: {
+        startDate: { lte: now },
+        endDate: { gte: now },
+        memberId: locals.user?.memberId,
+      },
+      select: {
+        id: true,
+      },
+    }) !== null;
```
danieladugyan, line 125: @alfredgrip Is this really correct? It looks to me like we're checking if `promise !== null`, since we're not awaiting the promise here but rather with Promise.allSettled.

alfredgrip: Great catch! Fixed in #536

PR #556, opened 2024-11-05
Fix switching languages when writing a post clears all content

### LINKED ISSUE #551: Switching languages when writing a post clears all content
### Description
If you're writing a news post and accidentally click the language switcher, it clears all of the content you have written. Even if you cancel the prompt to navigate away, it still clears all content...

I would expect the content to still be there if I accidentally navigate away from the page. At the very least, cancelling the navigation should absolutely make sure the content is still there.

Image

### Steps to reproduce
1. Go to https://www.dsek.se/news/create
2. Input some content
3. Click the language switcher
4. Cancel the navigation
5. The content has still disappeared

RufusKoggRojder:
Fix #551 by making the language switcher check if the language actually changed before it invalidates the page.


danieladugyan (approved):
LGTM! ✨

PR #578, opened 2024-11-12
Add feature flags

### LINKED ISSUE #573: Add a simple feature flag system
### Description
We often develop large features on a single long-lived feature branch. This discourages continuous integration and instead often leads to an annoying merge process once the feature has been developed. A feature flag system could help with this - a way to temporarily wrap new features with some conditional logic that checks whether the feature should be visible to users. That way we could merge incrementally without exposing unfinished features.

### Solution
Implement feature flagging. Maybe all feature flags should be enabled by default in development and disabled by default in production? Ideally, there would be a way for us to enable features for ourselves in production as well (but not for all users) so we can try them out.

Isak-Kallini:
closes #573

### DIFF
--- a/src/lib/utils/featureFlag.ts
+++ b/src/lib/utils/featureFlag.ts
@@ -1,0 +1,24 @@
+export const featureFlags: string[] = [];
+
+export const isFeatureFlagEnabled = (flag: string) => {
+  if (!featureFlags.includes(flag)) {
+    return false;
+  }
+  const value = localStorage.getItem(flag);
+  if (value === null || value === "false") {
+    return false;
+  }
+  return true;
+};
+
+export const setDefaultFlags = () => {
+  featureFlags.forEach((f) => localStorage.setItem(f, "false"));
+};
+
+export const setFeatureFlag = (flag: string, value: boolean) => {
+  if (!featureFlags.includes(flag)) {
+    return false;
+  }
+  localStorage.setItem(flag, value.toString());
+  return true;
+};
--- a/src/routes/(app)/admin/debug/+page.svelte
+++ b/src/routes/(app)/admin/debug/+page.svelte
@@ -2,56 +2,94 @@
   import { version } from "$app/environment";
   import { page } from "$app/stores";
   import { isAuthorized } from "$lib/utils/authorization";
+  import {
+    featureFlags,
+    isFeatureFlagEnabled,
+    setFeatureFlag,
+  } from "$lib/utils/featureFlag";
+  import { onMount } from "svelte";
 
   export let data;
   $: user = data.user;
   $: policies = user.policies.toSorted();
+  const flags = new Map<string, boolean>();
+  onMount(() => {
+    featureFlags.forEach((f) => {
+      flags.set(f, isFeatureFlagEnabled(f));
+    });
+  });
 </script>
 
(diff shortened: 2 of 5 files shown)


src/routes/+layout.svelte
```diff
@@ -6,13 +6,19 @@
...
   export let data;
 
   $: (() => {
     const locale = languageTag();
     dayjs.locale(locale);
   })();
+
+  onMount(() => {
+    setDefaultFlags();
+  });
```
alfredgrip, line 21: Why do we always set the flags to false when loading? This causes all flags to become false on all page reloads.

src/routes/(app)/admin/debug/+page.svelte
```diff
@@ -2,56 +2,94 @@
...
+  <div>
+    <section>
+      <h1 class="text-lg font-semibold">Feature flags</h1>
+      {#each featureFlags as flag}
+        <div class="flex gap-5">
+          {flag}
+          <input
+            type="checkbox"
+            class="toggle"
+            checked={flags.get(flag)}
```
alfredgrip, line 86: The state and the component isn't in sync. If you disable `setDefaultFlags();` (as described in my other comment), the `checked` is not working as intended

src/lib/utils/featureFlag.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export const setDefaultFlags = () => {
+  featureFlags.forEach((f) => localStorage.setItem(f, "false"));
+};
+
+export const setFeatureFlag = (flag: string, value: boolean) => {
+  if (!featureFlags.includes(flag)) {
+    return false;
+  }
+  localStorage.setItem(flag, value.toString());
```
alfredgrip, line 22: Maybe instead of having many different storage entries for each feature flag, perhaps we can wrap it in an object and JSON serialize/parse it? But this is opinionated and skip it if it's too big of a hassle

alfredgrip (commented):
Nice work overall, main functionallity seems to be there! I have some comments though

src/routes/+layout.svelte
Isak-Kallini: Oh yes that shouldn't be there, forgot to remove after testing some stuff

src/routes/(app)/admin/debug/+page.svelte
Isak-Kallini: Good catch!

src/lib/utils/featureFlag.ts
Isak-Kallini: I think it's worth keeping as is, since this way it's easier to enable/disable flags manually through dev tools if needed

alfredgrip (approved):
LGTM! Should we add a link to this in the navbar? Right now you manually have to enter the URL (I don't know if this is by design or not). I also pushed a minor styling change, changing the active toggles to pink

src/lib/utils/featureFlag.ts
```diff
@@ -0,0 +1,20 @@
+export const featureFlags: string[] = [];
+
+export const isFeatureFlagEnabled = (flag: string) => {
```
danieladugyan, line 3: If we want to avoid the issue of having to wrap this inside an `onMount()`, I think we could just `return false` from this function if `browser` is false.

PR #586, opened 2024-11-18
Fix #538 "Bookings for styrelserummet should show a warning"

### LINKED ISSUE #538: Bookings for styrelserummet should show a warning
### Description
The booking feature on the old dsek.se page used to show a warning to members who booked "styrelserummet", since this space is reserved for board members between 11:30 and 13:00 (not any other time).

Image
https://old.dsek.se/sektionen/km/boka.php

If someone checks the checkbox for "styrelserummet" when creating a booking, show an [alert](https://daisyui.com/components/alert/) containing a similar message to the one above.

There's no other good way to convey this information to bookers, except for handling it on a case-by-case basis.

RufusKoggRojder:
Fixed #538 by adding a warning when booking the council room.


danieladugyan (changes requested):
Other than my minor comment it looks good! 🌟

alfredgrip:
I took the freedom to change "council" to "board", since this is how we usually translate it. We use "council" when refering to "råd" (such as "studierådet" and "trivselrådet"). Revert it if you object 😇

danieladugyan (approved):
Thanks for the fix - LGTM!

PR #589, opened 2024-11-18
Add videos to articles

### LINKED ISSUE #479: Add support for videos in news articles
We have support for images but supporting a video would be nice. We have sponsors who wants to send out a video to everyone. The video should be viewable directly from the feed without having to open the article.

RufusKoggRojder:
Fixed #479. Allows the user to specify both a video and an image but will only show the video in the main article viewer.

### DIFF
--- a/src/routes/(app)/news/ArticleForm.svelte
+++ b/src/routes/(app)/news/ArticleForm.svelte
@@ -16,6 +16,7 @@
   export let allTags: Tag[];
   export let superform: SuperForm<ArticleSchema>;
   export let articleImage: string | undefined = undefined;
+  export let articleVideo: string | undefined = undefined;
 
   const { form, enhance, errors } = superform;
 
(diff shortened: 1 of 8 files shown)


src/routes/(app)/news/ArticleForm.svelte
```diff
@@ -52,6 +53,31 @@
...
+          new Uint8Array(result).reduce(
+            (data, byte) => data + String.fromCharCode(byte),
+            "",
+          ),
+        );
+      } else {
+        articleVideo = result;
+      }
+    };
+  };
```
danieladugyan, line 80: I'm not really sure how this code works 😅, but I have a few thoughts:
1. For one, it's very similar to the code for handling images. Maybe it could just be one function?
2. Does it work for large videos? As far as I can tell, the function converts the video to a long [base64](https://developer.mozilla.org/en-US/docs/Web/URI/Schemes/data) string and that feels a bit weird to me - is there any limit to how long a base64 string can be?

danieladugyan (changes requested):
I realised when looking though the code that this needs some more work until it will work with uploaded articles. The preview system works differently than when articles are actually uploaded.

Take a look at how image uploading is handled in `src/lib/news/server/actions.ts` and let me know if you have any questions! You can get access to the development file server if you want to try it out "for real".

RufusKoggRojder:
Video hosting became very troublesome due to upload size limitations. We may still try supporting YouTube videos.

Issue #579, opened 2024-11-12
All members should have access to iDét


danieladugyan:
### Description

Door access policies for `*` don't work at the moment. We seem to determine which members should have access based on first checking which roles should have access, which doesn't work for `*` where we want to find all members, even if they don't have any roles.

### Steps to reproduce

https://www.dsek.se/salto/idet only shows members which have a role/have been given direct access.

PR #590, opened 2024-11-19
Added support for wildcard roles in door access policies

Fiery-132:
Fix for #579. Wildcard roles will now fetch all members with class years up to 10 years back.

### DIFF
--- a/src/routes/(app)/salto/[door]/+server.ts
+++ b/src/routes/(app)/salto/[door]/+server.ts
@@ -100,16 +100,10 @@ export const GET: RequestHandler = async ({ params }) => {
     const { studentIdsBanned } = parseDoorBanPolicies(policies);
 
     const studentsFromWildcard = positionIds.includes("*")
-      ? await authorizedPrismaClient.member
+      ? authorizedPrismaClient.member
           .findMany({
-            where: {
-              classYear: {
-                gte: new Date().getFullYear() - 10,
-              },
-            },
-            select: {
-              studentId: true,
-            },
+            where: { classYear: { gte: new Date().getFullYear() - 10 } },
+            select: { studentId: true },
           })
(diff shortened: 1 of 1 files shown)


src/routes/(app)/salto/[door]/+server.ts
Comment (danieladugyan) [inline src/routes/(app)/salto/[door]/+server.ts:1]: I would make a few small changes:
1. Remove the initial await from `member.findMany()` since none of the following database queries depend on the result. This allows the subsequent database queries to run without waiting for the result of the first query. Then we can await the result later.
2. This is a minor style change, but I like [compressing objects](https://prettier.io/docs/en/rationale.html#multi-line-objects) that just contain one property (or few) so they take up fewer lines.

```ts
    const studentsFromWildcard = positionIds.includes("*")
      ? authorizedPrismaClient.member
          .findMany({
            where: { classYear: { gte: new Date().getFullYear() - 10 } },
            select: { studentId: true },
          })
          .then((members) =>
            members
              .map((member) => member.studentId)
              .filter((id): id is string => id !== null)
          )
      : [];

    ...

    /** students with current access to the door who are not banned */
    const allowedStudents = [
      ...studentIds,
      ...studentsFromPositions,
      ...(await studentsFromWildcard),
    ].filter((studentId) => !bannedStudents.has(studentId));
```

danieladugyan (changes requested):
Other than that, nice fix! 👍

src/routes/(app)/salto/[door]/+server.ts
Comment (Fiery-132) [reply src/routes/(app)/salto/[door]/+server.ts:1]: I agree and they should be implemented in the new commit i just made. As for `2.`, I didn't know Prettier doesn't collapse multiline code so thanks for the link! 😊

danieladugyan (approved):

PR #592, opened 2024-11-20
Svelte 5 upgrade and fixed immediate breaking changes

stagrim:
Mostly solved form inside form issue by moving the dialog outside the outside form, and submit the form using the form attribute on the button elements


stagrim:
Migrated to new format; hopefully everything works as well as it did previously now! We have a lot of illegal self closing tags though...

stagrim:
Went through and manually fixed most lint issues. Svelte 5 seems to introduce many type errors related to props with components though. Don't know how to fix those

stagrim:
Seems like I have managed to break the first page for those who are not logged in. I am not sure how to fix it

src/lib/components/forms/FormSubmitButton.svelte
```diff
@@ -10,12 +10,14 @@
...
   let clazz: string | undefined = undefined;
   export { clazz as class };
+  export let form: string | undefined = undefined;
   const { delayed } = superform;
 </script>
 
 <LoadingButton
   class={clazz}
   type="submit"
+  {form}
```
Isak-Kallini, line 20: What does this do?

Isak-Kallini (approved):
One small comment, looks good otherwise!

src/lib/components/forms/FormSubmitButton.svelte
stagrim: image

Issue #532, opened 2024-10-23
Cafe page missing translation and logo


fgren:
Cafemästeriet page is missing the Swedish translation and its logo. Perhaps related to #520 (the underlying issue or the fix?)
(Also, the volunteer benefits link in the text should be changed to /info/benefits)

danieladugyan:
I think I have a copy of the content stored somewhere - I'll try and find it and restore the content. As for why it breaks, I mentioned in https://github.com/Dsek-LTH/web/issues/519 that I think it's still broken.

danieladugyan:
Fixed now. Note to self, using random private Discord servers to store content backups make it quite difficult to find the backup... 😅

PR #595, opened 2024-11-20
Cafe buggy editing

### LINKED ISSUE #519: Editing cafe´s opening hours is buggy
1. The current opening hours are shown as placeholders in the input fields, instead of being shown as default values. This means that you can't easily edit the current opening hours.
2. If you submit a change, all other input fields on the entire page get cleared.

Steindt:
Might've solved issues in #519 and #532. Problem is there are 2 forms on the same page, when submitting one of them, the other might reset, ending up with empty input fields. I've added an id to one of the forms and I can't replicate the issue anymore, or find any more.

Closes #519 and #532

### DIFF
--- a/src/routes/(app)/committees/EditCommitteeForm.svelte
+++ b/src/routes/(app)/committees/EditCommitteeForm.svelte
@@ -19,6 +19,7 @@
   const superform = superForm(formData, {
     validators: zodClient(updateSchema),
     resetForm: false,
+    id: "generic",
   });
   const { form, enhance } = superform;
 </script>


fgren (approved):
LGTM

danieladugyan:
It might work - I'm not sure. There's quite a lot of weird stuff going on on the café page:
1. The opening hours use plain forms instead of Superforms. 
2. There's one form for every day of the opening hours instead of just one form for the entire component.
3. Form id's [shouldn't be needed if we use Superform's enhance method](https://superforms.rocks/concepts/multiple-forms), which we do for the superform (but obviously not for the plain form).

danieladugyan (commented):
It might work - I'm not sure. There's quite a lot of weird stuff going on on the café page:

The opening hours use plain forms instead of Superforms.
There's one form for every day of the opening hours instead of just one form for the entire component.
Form id's [shouldn't be needed if we use Superform's enhance method](https://superforms.rocks/concepts/multiple-forms), which we do for the superform (but obviously not for the plain form).

Steindt:
True, it shouldn't be necessary as one is superforms and the other isn't, and the superforms uses use:enhance. But havn't been able to reproduce the bug again since I've set the id.  If anyone is able to get the bug again, then it clearly doesn't work and we might have to look at changing how the forms are set up.

Issue #583, opened 2024-11-16
Long member nicknames look terrible


danieladugyan:
### Description

![image](https://github.com/user-attachments/assets/9fced806-adff-479f-9a8a-e36da6af0ac6)
![image](https://github.com/user-attachments/assets/1527acdc-5f45-4bc9-bd3e-900f211238a4)
![image](https://github.com/user-attachments/assets/5746e6d6-3286-4a83-8174-bdd39ae96c73)

### Steps to reproduce

https://www.dsek.se/committees/infu?year=2024
https://www.dsek.se/members/al7718pi-s
https://www.dsek.se/members?programme=D&year=2022

### Further information

I did some initial investigation and sorted user's nickname's by length. Here are the results:

|  n  |
|-----|
| 248 |
| 137 |
|  98 |
|  85 |
|  51 |
|  51 |
|  48 |
|  45 |
|  44 |
|  40 |
|  39 |
|  39 |
|  39 |
|  36 |
| ...   |

PR #602, opened 2024-11-26
Fix: Long member nicknames look terrible

ekvanox:
Fixes issue #583: https://github.com/Dsek-LTH/web/issues/583

**Before**

bild

**Now**

bild


Joel-Liden (approved):
These look like some great changes to how text is displayed!

PR #608, opened 2024-11-26
fix: data leak on home page

### LINKED ISSUE #550: Information leak in website source code
### Description

Information that I would expect to be somewhat private is currently accessible through the website's source code. For instance, it seems LOTS of rows from our member database is simply inlined into the source code. The screenshot below is a very small excerpt from what you can see in the source code.

image

### Steps to reproduce

Step 1. Go to [dsek.se](https://www.dsek.se/home)
Step 2. Right-click and choose "view source".
Step 3. Scroll down to the bottom script and see the exposed data.

### Further information

SvelteKit has a [feature](https://svelte.dev/docs/kit/load#Making-fetch-requests) where *"During server-side rendering, the response will be captured and inlined into the rendered HTML"* which seems related to this. However, that's related to making `fetch()` calls so I'm not sure what's going on here.

Isak-Kallini:
closes #550
This pr doesn't really fix the issue (especially since our api exposes all the data anyways). The problem is that we return way to much data, we need some kind of solution for protecting private data like food preference and mail. I will make a new issue discussing it more.

### DIFF
--- a/src/lib/components/socials/AuthorAvatars.svelte
+++ b/src/lib/components/socials/AuthorAvatars.svelte
@@ -2,11 +2,11 @@
   import CustomAuthorImage from "$lib/components/socials/CustomAuthorImage.svelte";
   import DsekAvatar from "$lib/components/socials/DsekAvatar.svelte";
   import MemberImage from "$lib/components/socials/MemberImage.svelte";
-  import type { Author, CustomAuthor, Member } from "@prisma/client";
+  import type { Author, CustomAuthor } from "@prisma/client";
 
   export let authors: Array<
     Author & {
-      member: Member;
+      member: { picturePath: string | null };
       customAuthor: CustomAuthor | null;
     }
   >;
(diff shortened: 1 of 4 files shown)


Macludde:
I don't think email is private data? The idea is to display it on someone's profile? I guess the access policy could be "_", i.e. logged in such that crawlers won't find it

Isak-Kallini:
> I don't think email is private data? The idea is to display it on someone's profile? I guess the access policy could be "_", i.e. logged in such that crawlers won't find it

@danieladugyan  pointed out that alumns can have their private emails instead of the user.dsek one, which probably shouldn't be public

Macludde:
> > I don't think email is private data? The idea is to display it on someone's profile? I guess the access policy could be "_", i.e. logged in such that crawlers won't find it
> 
> 
> 
> @danieladugyan  pointed out that alumns can have their private emails instead of the user.dsek one, which probably shouldn't be public

Still, it should be visible to other members. The whole idea of adding your email is to display it on your profile, I think it's a great way to contact an guild member.

On old.dsek.se, this is how it worked, and that's the terms the alumni "agreed to" when giving up their email.

We can hide it if not logged in.

Isak-Kallini:
Solved with basically the same changes in #624

Issue #510, opened 2024-10-09
Cannot add user.dsek.se mail to special receivers


johannesnydahl:
When editing a mail alias, one can not add new special receivers using the user.dsek.se domain, however, it does work for the dsek.se domain. 

![Recording](https://github.com/user-attachments/assets/c34fb7af-89e9-48eb-9c26-6973f9e67fb9)

PR #612, opened 2024-11-26
fix: allow user.dsek.se

ekvanox:
Allow emails from `user.dsek.se`

Fixes: https://github.com/Dsek-LTH/web/issues/510


Joel-Liden (approved):
yeah, makes sense

PR #619, opened 2024-11-26
fix: Editing English position names does not work

ekvanox:
Fixes: https://github.com/orgs/Dsek-LTH/projects/7/views/8?pane=issue&itemId=81630562&issue=Dsek-LTH%7Cweb%7C496


Macludde (approved):
LGTM. However, seeing this makes me think we should decide on one way to handle editing of english texts.

Sometimes we have two fields, sometimes we use "tabs", sometimes (like here) we replace the content the user sees. This is however out-of-scope for this pr.

ekvanox:
yes, i agree with that

PR #626, opened 2024-12-05
Upgrade to Svelte 5 again

danieladugyan:
TO-DO: Test that the page still works!

#592 broke some forms so it has been temporarily reverted. The changes are kept on this branch so that they can be merged again once the issues are resolved.


Isak-Kallini:
I've clicked around and tried most forms and it seems to be working now, it's very possible that I've missed something so it would be good if someone else clicked around a bit as well

o-korpi (approved):
lgtm :+1: 
Tried most forms I could find, nothing appears to be broken.

PR #611, opened 2024-11-26
Add Meilisearch endpoint

alfredgrip:
This pull request introduces the integration of Meilisearch into the project to enhance search capabilities. The main changes include setting up Meilisearch, configuring the environment, and implementing search-related functionality.

### Environment and Configuration:
* [`.env`](diffhunk://#diff-e9cbb0224c4a3d23a6019ba557e0cd568c1ad5e1582ff1e335fb7d99b7a1055dR68-R72): Added Meilisearch configuration variables `MEILISEARCH_HOST` and `MEILISEARCH_MASTER_KEY`.
* [`docker-compose.yml`](diffhunk://#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R16-R28): Added a Meilisearch service configuration.
* [`dev/setup_db.sh`](diffhunk://#diff-7d07b10677b7eac294f69762d04254cc41245e3f84e1349abec5acb71690277eL2-R2): Updated the script to use `docker compose` instead of `docker run` for setting up the database.

### Documentation:
* [`docs/reference/external-systems.md`](diffhunk://#diff-4a4d4e85f1e7fe035a9768a897218cebe1fc039b85c1a664e79007070058460fR24-R27): Documented the integration of Meilisearch and its purpose.

### Dependency Management:
* [`package.json`](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R127): Added the `meilisearch` package as a dependency.
* [`pnpm-lock.yaml`](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbR109-R111): Updated to include the `meilisearch` package. [[1]](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbR109-R111) [[2]](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbR4088-R4090) [[3]](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbR10588-R10589)

### Search Functionality:
[...]

PR #627, opened 2024-12-06
Search UI

alfredgrip:
This PR rewrites search in frontend so that it uses the Meilisearch endpoint introduced in #611 

It also adds a "advanced search" that allows filtering search on different indexes

It also fixes some minor TS misses that slipped through in the #611 PR

To try out this PR, please ensure that Meilisearch is up and running on your Docker (rerun the `dev/setup_db.sh` script) and then head to `localhost:.../admin/debug` and press `Sync with Meilisearch`


src/routes/(app)/search/+page.server.ts
```diff
@@ -1,14 +1,45 @@
...
+
+    const url = new URL("/api/search", event.request.url);
+    url.searchParams.set("query", query);
+    url.searchParams.set("indexes", JSON.stringify(indexes));
+    url.searchParams.set("language", language);
+    const response = await event.fetch(url);
+    if (!response.ok) {
+      // silently fail
+      return;
+    }
```
Isak-Kallini, line 34: Who no error handling? We should at least log errors so we can investigate any problems in the future

src/routes/(app)/search/+page.server.ts
```diff
@@ -1,14 +1,45 @@
...
+    const includeEvents = data.get("events") === "on";
+    const includeArticles = data.get("articles") === "on";
+    const includePositions = data.get("positions") === "on";
+    const includeSongs = data.get("songs") === "on";
+    const indexes = [];
+    if (includeMembers) indexes.push("members");
+    if (includeEvents) indexes.push("events");
+    if (includeArticles) indexes.push("articles");
+    if (includePositions) indexes.push("positions");
+    if (includeSongs) indexes.push("songs");
```
Isak-Kallini, line 22: Nitpick but this could be shorter, something like
```
const indexes = [
  "members",
  "events",
  "articles",
  "positions",
  "songs",
].filter((v) => data.get(v) === "on");
```

src/lib/components/search/ArticleSearchResult.svelte
```diff
@@ -0,0 +1,32 @@
...
+    <div class="avatar aspect-square w-8 overflow-hidden rounded-full">
+      <span class="i-mdi-newspaper text-2xl"></span>
+    </div>
+    <div>
+      <h4 class="text-balance">
+        {#if languageTag() === "en"}
+          {article.headerEn}
+        {:else}
+          {article.header}
+        {/if}
```
Isak-Kallini, line 21: Could this be handled in the backend instead so we can avoid this kind of if else everywhere?

src/lib/components/search/GlobalSearch.svelte
```diff
@@ -0,0 +1,268 @@
...
+    isOpen = true;
+    document.body.style.overflow = "hidden";
+  }
+
+  function close() {
+    dialog.close();
+    isOpen = false;
+    document.body.style.overflow = "auto";
+    // input = "";
+    // results = [];
```
Isak-Kallini, line 58: Remove?

Isak-Kallini (changes requested):
Search in english doesn't seem to be working, for example:
The bottom article comes up when searching for test
image
But when searching for its title it doesn't show up:
image
This is with language set to english, the article also comes up if i search for it's swedish title

Also if I press backspace while the input is empty the spinner appears until I type something

alfredgrip:
> Search in english doesn't seem to be working

I found the issue for this. The `/search` page does a fetch to the api endpoint (which in turn makes a request to Meilisearch). When the server on the search page gets a request, the `event.locals.language` gets populated to the correct value by `hooks.server.ts`. But when the `/search` does a fetch to the API endpoint, the server there checks for `event.locals.language` but it defaults to swedish.
Tried some console logging and found this when trying to search with language set to `en`:
```
search param language is: en      // query parameter
search locals language is: en     // event.locals.language
API param language is: en         // query parameter
API locals language is: sv        // event.locals.language
```
So either the `hooks.server.ts` is faulty, or the `/search` server needs to do the fetch some other way. 

@danieladugyan Do you have any thoughts about this?

alfredgrip:
> I found the issue for this. The `/search` page does a fetch to the api endpoint (which in turn makes a request to Meilisearch). When the server on the search page gets a request, the `event.locals.language` gets populated to the correct value by `hooks.server.ts`. But when the `/search` does a fetch to the API endpoint, the server there checks for `event.locals.language` but it defaults to swedish. Tried some console logging and found this when trying to search with language set to `en`:
> 
> ```
> search param language is: en      // query parameter
> search locals language is: en     // event.locals.language
> API param language is: en         // query parameter
> API locals language is: sv        // event.locals.language
> ```
> 
> So either the `hooks.server.ts` is faulty, or the `/search` server needs to do the fetch some other way. When the API page instead gets language from the query parameter, search in english works as intended.

I found that this can simply be fixed if the `/search` server does a request to `/en/api/search` if `locals.language == "en"`. I'm not a big fan of this, however. Having two "different" API endpoints that does the same, but which could easily be merged in to a single one by having a query parameter i.e. `?language="en"`, seems weird. I think it makes more sense to have a single one that accepts the language query parameter. What are your thoughts @Isak-Kallini @danieladugyan ?

danieladugyan:
> I found that this can simply be fixed if the /search server does a request to /en/api/search if locals.language == "en". I'm not a big fan of this, however. Having two "different" API endpoints that does the same, but which could easily be merged in to a single one by having a query parameter i.e. ?language="en", seems weird. I think it makes more sense to have a single one that accepts the language query parameter. What are your thoughts @Isak-Kallini @danieladugyan ?

Sorry, that's my bad for introducing the bug. Anyways, that's exactly how I think it should work. It's consistent with how language is handled everywhere else in the app.
1. The request hits the server. 
2. `paraglide.js` looks at the URL and check if it has a language prefix, i.e `/en`, and sets the language in `event.locals` accordingly.
3. Route handlers use `event.locals.language` as needed - the Prisma `translationExtension` doing this automatically is enough in most cases.

In summary, if we're on the Swedish site, fetch normally, otherwise fetch with `'en'`  prepended. Ideally this would be done automatically by the library, but maybe we could make a `fetch` wrapper?

alfredgrip:
I think it's weird though, specifics sent to an API endpoint should IMO be in the query parameters, not extracted from the URL. But I guess this works.

Just pushed some changes that fixes the english search and also handle translation mapping.


> but maybe we could make a fetch wrapper?

So we would write a wrapper for the already existing `event.fetch` wrapper from SvelteKit? 🫠

src/routes/(app)/search/+page.server.ts
Isak-Kallini: Could we either just log the error server side, or add error handling client side? Otherwise we just get the generic 500 internal server error which is not very nice

Isak-Kallini:
I think that solution looks fine since, as Daniel says, that's how language is handled everywhere else.

src/routes/(app)/search/+page.server.ts
alfredgrip: Fixed in latest push!

alfredgrip:
> I think that solution looks fine since, as Daniel says, that's how language is handled everywhere else. But should maybe use `i18n.resolveRoute('/api/search')`?

Didn't know that was a thing, but that's a nice way to handle it I guess!

src/lib/components/search/ArticleSearchResult.svelte
```diff
@@ -0,0 +1,24 @@
+<script lang="ts">
+  import type { ArticleSearchReturnAttributes } from "$lib/search/searchTypes";
+  export let article: ArticleSearchReturnAttributes;
+</script>
+
+<li>
+  <a
+    href={"/news/" + article.slug}
+    class="border border-transparent focus:border-primary"
+    id=".search-result"
```
danieladugyan, line 10: What does this id do? And why this it start with a period? AFAIK id's are supposed to be globally unique.

alfredgrip: This is a part that enables keyboard navigation with the search results, used in `captureListItems()`
```ts
listItems = Array.from(formElement?.querySelectorAll("a")).filter(
  (a) => a.id === ".search-result",
);
```
I agree that this isn't idiomatic, just changed it in my latest commit

Issue #491, opened 2024-09-19
Search feature for meeting documents


alfredgrip:
Would be fun and useful if there was a fuzzy search feature for meeting documents. Sometimes you might want to find a specific motion but can't remember which meeting it was brought up on. I think since basically all our documents are LaTeX PDFs, there probably exist some tool that allows for indexing and searching amongst them

alfredgrip:
Idea so far: for all PDFs, generate text files with their content using Poppler which provides `pdftotext`. Connect PDFs with the text files and provide some search functionality for the text files, but return PDF results.

https://poppler.freedesktop.org/
https://www.npmjs.com/package/node-poppler

alfredgrip:
@01ste02 I vaguely remember you had a script that checked how many times your name was mentioned in guild documents, how did that work?

01ste02:
I found an old message on discord containing this script:

```bash
#!/bin/bash
PERSON="Axel Svensson"
LATEST_MEETING=26

# Usage: ./filename.sh "Namn" SiffraFörSistaStyrelsemöte

if [ $# -eq 1 ]
    then
        if [[ "$1" =~ ^-?[0-9]+$ ]]; then
        LATEST_MEETING=$1
    else    
            PERSON=$1
    fi
elif [ $# -eq 2 ]
        then
        if [[ "$1" =~ ^-?[0-9]+$ ]]; then
                LATEST_MEETING=$1
                PERSON=$2
        else
                PERSON=$1
                LATEST_MEETING=$2
        fi
fi


for fix in ".pdf" "_2.pdf"; do 
    for i in $(seq -f "%02g" 1 ${LATEST_MEETING}); do 
        cd /tmp
        #echo "https://minio.api.dsek.se/documents/public/2023/S${i}/protokoll_S${i}_2023${fix}"
        wget -q https://minio.api.dsek.se/documents/public/2023/S${i}/protokoll_S${i}_2023${fix} 2> /dev/null
        pdftotext -l 1 /tmp/protokoll_S${i}_2023${fix} 2> /dev/null
        rm /tmp/protokoll_S${i}_2023${fix} 2> /dev/null
    done
done
echo "${PERSON} har närvarat på ca $(grep -lrnw "${PERSON}" /tmp/protokoll_S* | wc -l) styrelsemöten" 2> /dev/null
echo "Dessa möten har ${PERSON} förmodligen närvarat på:"
[...]

PR #636, opened 2024-12-10
Add a pdf to text service

alfredgrip:
This PR adds a separate web server that can convert a PDF to text, and respond with the text. Not sure this is how we should do it, but it works.

This is a step in the process to provide a search feature for documents #491 
As mentioned in #491 , I didn't have much success with Apache Tika, so I went the Poppler way instead.

I guess @stagrim and @danieladugyan might have opinions about this.


danieladugyan (approved):
Cool! ✨ I'm merging this with some small changes in 4e3cfe3eb14bcb834af9755f2817a6d6ab3f5df9.

PR #639, opened 2024-12-10
add year above the "nolla in group" section

lego-eden:


Joel-Liden (approved):
Looks like a good change

PR #650, opened 2024-12-11
Add election editing

fgren:
Adds functionality to create and edit elections.

### DIFF
--- a/src/routes/(app)/elections/+page.svelte
+++ b/src/routes/(app)/elections/+page.svelte
@@ -4,11 +4,18 @@
   import * as m from "$paraglide/messages";
   import MarkdownBody from "$lib/components/MarkdownBody.svelte";
   import { languageTag } from "$paraglide/runtime";
+  import { isAuthorized } from "$lib/utils/authorization";
+  import apiNames from "$lib/utils/apiNames";
   export let data: PageData;
 </script>
 
-<PageHeader title={m.openElections()} />
-<!-- TODO: make this editable by board members with Markdown page -->
+<div class="flex flex-row">
+  <PageHeader title={m.openElections()} />
+  {#if isAuthorized(apiNames.ELECTION.CREATE, data.user)}
+    <a href={"/elections/new"} class="btn btn-primary ml-auto">+ Nytt val</a>
+  {/if}
+</div>
+
 <section class="mb-5 space-y-5">
   <p>
     {m.elections_description()}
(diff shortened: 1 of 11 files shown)


src/routes/(app)/elections/+page.svelte
```diff
@@ -12,7 +12,7 @@
 <div class="flex flex-row">
   <PageHeader title={m.openElections()} />
   {#if isAuthorized(apiNames.ELECTION.CREATE, data.user)}
-    <a href={"/elections/new"} class="btn btn-primary ml-auto">+ Nytt val</a>
+    <a href="/elections/create" class="btn btn-primary ml-auto">+ Nytt val</a>
```
danieladugyan, line 16: I changed this to create since that's more consistent with our other pages (except for one for some reason...).

src/routes/(app)/elections/+page.svelte
```diff
@@ -29,8 +29,9 @@
         {#if isAuthorized(apiNames.ELECTION.UPDATE, data.user)}
           <a
             href={"/elections/" + election.id + "/edit"}
-            class="btn btn-secondary btn-sm ml-auto">Redigera</a
-          >
+            class="btn btn-outline btn-sm ml-auto"
```
danieladugyan, line 33: We have a lot of purple (`btn-secondary`) buttons on the page, but I'm not a big fan of them so I changed this one to a more neutral color.

I'm not really an expert on UI design (clearly... 😄) but, colors tend to draw attention and usually signify some kind of main action we want the user to take. Editing is more of an optional action and thus doesn't need to be highlighted.

src/routes/(app)/elections/ElectionEditor.svelte
```diff
@@ -4,18 +4,16 @@
...
-      markdown: string;
-      markdownEn: string | null;
-      link: string;
-      expiresAt: Date;
-      committeeId: string;
-    };
+    election: Pick<
+      Election,
+      "markdown" | "markdownEn" | "link" | "expiresAt" | "committeeId"
+    >;
```
danieladugyan, line 16: I changed to to use TypeScript's `Pick` utility type so that we can reuse `Election` type that already exists.

danieladugyan (approved):
I made some changes and left comments to explain them, but nothing of much importance really. The comments are just there if you're curious as to why I made the changes. Nice work! ⭐

As always, feel free to disagree 😄, otherwise we can go ahead and merge this.

PR #653, opened 2024-12-16
Search fixes

alfredgrip:
Since the new search was deployed, some issues have been spotted and this PR attempts to fix some things.

Main thing: timeout for each task is now 60 seconds, and adding documents is split up into different tasks with different timeouts


Isak-Kallini (approved):
LGTM! Nice with logging as well

PR #660, opened 2024-12-19
Reduce data sent to client during search

alfredgrip:
Slices string attributes (which aren't used as link) to 60 characters. Why 60? It's less but still gives a nice preview in the search results. If we were to slice strings that are used as links, we might return a borken link.

### DIFF
--- a/src/routes/(app)/api/search/+server.ts
+++ b/src/routes/(app)/api/search/+server.ts
@@ -78,10 +78,10 @@ export const GET: RequestHandler = async ({ url, locals }) => {
         }
 
         // Reduce the amount of data sent to the client
-        // all string values are capped at 60 characters
+        // all string values are truncated to 60 characters
         for (const key of Object.keys(hit)) {
           if (
-            // We musn't cap the attributes that are used as links
+            // We must not truncate attributes that are used as links
             !listOfattributesUsedAsLink.includes(key) &&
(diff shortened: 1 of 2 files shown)


src/routes/(app)/api/search/+server.ts
```diff
@@ -73,6 +76,19 @@ export const GET: RequestHandler = async ({ url, locals }) => {
           }
           hit = hitCopy;
         }
+
+        // Reduce the amount of data sent to the client
+        // all string values are capped at 60 characters
+        for (const key of Object.keys(hit)) {
+          if (
+            // We musn't cap the attributes that are used as links
```
danieladugyan, line 84:
```suggestion
            // We must not truncate attributes that are used as links
```

src/routes/(app)/api/search/+server.ts
```diff
@@ -73,6 +76,19 @@ export const GET: RequestHandler = async ({ url, locals }) => {
           }
           hit = hitCopy;
         }
+
+        // Reduce the amount of data sent to the client
+        // all string values are capped at 60 characters
```
danieladugyan, line 81:
```suggestion
        // all string values are truncated to 60 characters
```

danieladugyan (approved):
LGTM! (don't mind the comments, I'll just commit them myself)

Issue #661, opened 2024-12-20
Duplicate data in search


Isak-Kallini:
### Description

![image](https://github.com/user-attachments/assets/e71d7875-522d-453b-a395-68e14fcb8462)
I think this is because the daily sync is run in each instance at the same time (eg. 4 instances in the screenshot)

alfredgrip:
Is there any way to ensure only one instance runs this? @danieladugyan

alfredgrip:
I think this could have been solved if we didn't generate ids when putting data in meilisearch, but instead use our primary keys from Prisma. But this haven't worked for me, since I think some of the ids we use for some tables aren't compatible with Meilisearch. Meilisearch likes uuids, but for Positions we use like `dsek.infu.dwww.mdlm`, and I suspect that Meilisearch doens't like this form of ids.
If we didn't generate ids, Meilisearch would see that we are sending duplicate data and not create four different entries.

alfredgrip:
> I think this could have been solved if we didn't generate ids when putting data in meilisearch, but instead use our primary keys from Prisma. But this haven't worked for me, since I think some of the ids we use for some tables aren't compatible with Meilisearch. Meilisearch likes uuids, but for Positions we use like `dsek.infu.dwww.mdlm`, and I suspect that Meilisearch doens't like this form of ids. If we didn't generate ids, Meilisearch would see that we are sending duplicate data and not create four different entries.

Just confirmed this: 
> The document id must be an integer or a string. If the id is a string, it can only contain alphanumeric characters (a-z, A-Z, 0-9), hyphens (-), and underscores (_).

https://www.meilisearch.com/docs/learn/getting_started/primary_key#formatting-the-document-id
So it cannot contain dots `.`

alfredgrip:
Added a commit in #662 which should solve this temporarily, but ideally it should be run on only one instance.

alfredgrip:
Maybe something like this could solve this issue? 
https://greenydev.com/blog/pm2-cron-job-multiple-instances/#solution-1-setting-app-name-in-pm2-instances 
It should be relevant for all cron jobs we do, not only syncing with Meilisearch.

@danieladugyan

PR #662, opened 2024-12-20
Refactor search sync to try to reduce memory usage and hash ids

alfredgrip:
Title speaks for itself.

Idea is to divide data in batches and send it one at a time, this way the garbage collector should be able to drop objects since they go out of scope.

Current batch size is 1000, which I think should be fine.

Also hashes ids which should be a (temporary) fix to  #661

### DIFF
--- a/src/lib/search/searchHelpers.ts
+++ b/src/lib/search/searchHelpers.ts
@@ -11,8 +11,6 @@ import {
 } from "./searchTypes";
 import * as m from "$paraglide/messages";
 
-import crypto from "crypto";
-
 export type SearchIndex =
   (typeof availableSearchIndexes)[keyof typeof availableSearchIndexes];
 
@@ -141,7 +139,6 @@ export function mapIndexToMessage(index: SearchableIndex) {
 }
 
 /**
- * WARNING: This is non-reversible.
  * Meilisearch has constraints on the id field:
  * it can only contain alphanumeric characters (a-z, A-Z, 0-9), hyphens (-), and underscores (_).
  * So we need to convert the prisma id to a meilisearch id.
@@ -149,9 +146,5 @@ export function mapIndexToMessage(index: SearchableIndex) {
  * It contains dots (.) and possibly swedish characters (åäö).
  */
 export function prismaIdToMeiliId(id: string) {
-  const hashed = crypto.createHash("md5").update(id).digest("hex");
-  if (!hashed.match(/^[a-zA-Z0-9_-]+$/)) {
-    console.log(`Meilisearch: Id ${id} from Prisma is not compatible`);
-  }
-  return hashed;
+  return btoa(id).replace(/\+/g, "-").replace(/\//g, "_").replace(/=/g, "");
 }
--- a/src/lib/search/searchHelpers.ts
+++ b/src/lib/search/searchHelpers.ts
@@ -11,6 +11,8 @@ import {
 } from "./searchTypes";
 import * as m from "$paraglide/messages";
 
+import crypto from "crypto";
+
 export type SearchIndex =
   (typeof availableSearchIndexes)[keyof typeof availableSearchIndexes];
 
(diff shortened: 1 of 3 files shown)


src/lib/search/searchHelpers.ts
```diff
@@ -137,3 +139,19 @@ export function mapIndexToMessage(index: SearchableIndex) {
       return "";
   }
 }
+
+/**
+ * WARNING: This is non-reversible.
```
danieladugyan: > `WARNING: This is non-reversible.`

Does this matter? In any case, it's reversible now.

src/lib/search/sync.ts
```diff
@@ -32,243 +31,81 @@ const sync = async () => {
...
-  );
-  return tasks;
-}
+/**
+ * Meilisearch has constraints on the id field:
+ * it can only contain alphanumeric characters (a-z, A-Z, 0-9), hyphens (-), and underscores (_).
+ * So we need to convert the prisma id to a meilisearch id.
+ * As of now, the Positions table has a prisma id that is not compatible with meilisearch.
+ * (it contains dots)
+ */
```
danieladugyan, line 62: Is this comment supposed to be here as well?

src/lib/search/searchHelpers.ts
```diff
@@ -137,3 +137,14 @@ export function mapIndexToMessage(index: SearchableIndex) {
...
+/**
+ * Meilisearch has constraints on the id field:
+ * it can only contain alphanumeric characters (a-z, A-Z, 0-9), hyphens (-), and underscores (_).
+ * So we need to convert the prisma id to a meilisearch id.
+ * As of now, the Positions table has a prisma id that is not compatible with meilisearch.
+ * It contains dots (.) and possibly swedish characters (åäö).
+ */
+export function prismaIdToMeiliId(id: string) {
+  return btoa(id).replace(/\+/g, "-").replace(/\//g, "_").replace(/=/g, "");
+}
```
danieladugyan, line 150: I changed this function. MD5 seems like it could be [slow](https://stackoverflow.com/questions/6122571/simple-non-secure-hash-function-for-javascript#comment14096359_8831937) and something simple like base64 encoding should be fine.

danieladugyan:
There's a lot of code in `sync.ts` and `searchTypes.ts` , to be honest it's a bit hard for me to grasp it all😅

src/lib/search/searchHelpers.ts
alfredgrip: No it doesn't really matter, but just a heads up that we cannot convert back to Prisma ID given Meili ID. But doesn't matter now

src/lib/search/searchHelpers.ts
alfredgrip: Nice 👍 it doesn't need to be a fancy hashing funciton

src/lib/search/sync.ts
alfredgrip: It doesn't have to be there, it's more of a relic from when I used `uuid` instead. But since there is an explanatory comment documenting the converting funciton, this one could be removed

alfredgrip:
> There's a lot of code in `sync.ts` and `searchTypes.ts` , to be honest it's a bit hard for me to grasp it all😅

Yeah `searchTypes.ts` is a lot...
Essentially, for every index in Meili there is a type for:
1. What attributes is stored in Meili for that index
2. Which attributes can a user perform a search after
3. Which attributes are returned

Of course, all attributes that can be searched on, or are returned, must be stored in Meilisearch. It not as simple as doing a union of 2. and 3. to get 1. however, since some attributes (like`startDatetime` for events) are used purely for sorting and ranking purposes internally by Meili.

Then there are objects like `const memberMeilisearchConstants: MemberConstantsMeilisearch = {...` which wraps things related to an index in a single object. Here we can specify custom ranking and sorting rules for Meili, such as giving newer members a higher ranking, and tweak which typo tolerance is allowed.

I know that the file is full of types, but it is to prevent us developers from accidentally trying to e.g. do custom ranking rules on an attribute that isn't even stored in Meilisearch.

As for `sync.ts`, it's basically just dump the data and attributes defined in `searchTypes.ts` to Meilisearch, but do so in batches. When all the data is dumped, tweak the rankings based on the values and types also defined in `searchTypes.ts`.

danieladugyan:
> > There's a lot of code in `sync.ts` and `searchTypes.ts` , to be honest it's a bit hard for me to grasp it all😅
> 
> Yeah `searchTypes.ts` is a lot... Essentially, for every index in Meili there is a type for:
> 
> 1. What attributes is stored in Meili for that index
> 2. Which attributes can a user perform a search after
> 3. Which attributes are returned
> 
> Of course, all attributes that can be searched on, or are returned, must be stored in Meilisearch. It not as simple as doing a union of 2. and 3. to get 1. however, since some attributes (like`startDatetime` for events) are used purely for sorting and ranking purposes internally by Meili.
> 
> Then there are objects like `const memberMeilisearchConstants: MemberConstantsMeilisearch = {...` which wraps things related to an index in a single object. Here we can specify custom ranking and sorting rules for Meili, such as giving newer members a higher ranking, and tweak which typo tolerance is allowed.
> 
> I know that the file is full of types, but it is to prevent us developers from accidentally trying to e.g. do custom ranking rules on an attribute that isn't even stored in Meilisearch.
> 
> As for `sync.ts`, it's basically just dump the data and attributes defined in `searchTypes.ts` to Meilisearch, but do so in batches. When all the data is dumped, tweak the rankings based on the values and types also defined in `searchTypes.ts`.

Thanks! I added that as a comment to the top of the file since it helped a lot.

PR #685, opened 2025-01-19
show event link

### LINKED ISSUE #684: Show link in event
### Description

When creating or editing an event on the website one can add a link, I'm guessing to potential fb-events or meeting notices. Just as the event location until recently, this is not displayed on the actual event.

_No response_

_No response_

anahitachavan:
closes #684

### DIFF
--- a/src/routes/(app)/events/SmallEventCard.svelte
+++ b/src/routes/(app)/events/SmallEventCard.svelte
@@ -89,6 +89,21 @@
           </div>
         </section>
       {/if}
+      {#if event.link}
+        <section class="my-2">
+          <div class="flex items-center gap-2">
+            <span class="i-mdi-link text-lg"></span>
+            <a
+              href={event.link}
+              class="text-base leading-none hover:underline"
+              target="_blank"
+              rel="noopener noreferrer"
+            >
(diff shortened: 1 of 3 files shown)


stagrim:
Do we want the link to only display the text "link", or should it display the href to make it obvious where it links to? The positions spells it out instead of linking to google maps and displaying something like "position". Is this something we want to do or would that be too much information? Could perhaps truncate long links when displayed to avoid ugly overflow then.

Isak-Kallini:
> maybe DWWW-ansvarig @Isak-Kallini should have the final opinion?

I think either can work, what do you think?

src/routes/(app)/events/SmallEventCard.svelte
```diff
@@ -89,6 +89,21 @@
           </div>
         </section>
       {/if}
+      {#if event.link}
+        <section class="my-2">
+          <div>
+            <span class="i-mdi-link text-lg"></span>
```
Isak-Kallini, line 95: nit: icon is not centered vertically with the text

Isak-Kallini (approved):
lgtm, but conflicts need to be resolved

PR #697, opened 2025-01-30
improve keycloak sync

Isak-Kallini:
Adds a last_synced field to the mandates table and checks if a mandate has started or ended since the last time it synced and adds/deletes the user from the keycloak group.


Isak-Kallini:
There are currently 6432 entries in the mandates table, and it will just keep increasing so I think it's not sustainable to sync everything. 
Considering we already have a lot of problems with them drifting out of sync I don't see how my changes would make it worse (hopefully it makes it better). But it's definitely something that should be discussed more

fgren (approved):

PR #721, opened 2025-02-10
Fix data being leaked due to app page title

Macludde:


Joel-Liden (approved):
Looks like it works, as I still see the page title. I can't say if it fixes data leak _but I assume that's been fixed._

PR #728, opened 2025-02-17
Fixed language "randomly" switching

### LINKED ISSUE #575: Website randomly switches language to English
### Description

When I use the dsek.se website, it sometimes switches language from Swedish to English when I navigate between pages.

### Steps to reproduce

Absolutely no idea. Click around on the website for a bit? Good luck!

### Further information

_No response_

Fiery-132:
This should fix #575. Now sets the language tag in `hooks` instead of `layout.server`.

### DIFF
--- a/src/hooks.server.ts
+++ b/src/hooks.server.ts
@@ -5,7 +5,11 @@ import { i18n } from "$lib/utils/i18n";
 import { createMember } from "$lib/utils/member";
 import { redirect } from "$lib/utils/redirect";
 import { themes, type Theme } from "$lib/utils/themes";
-import { isAvailableLanguageTag, sourceLanguageTag } from "$paraglide/runtime";
+import {
+  isAvailableLanguageTag,
+  setLanguageTag,
+  sourceLanguageTag,
+} from "$paraglide/runtime";
 import Keycloak, { type KeycloakProfile } from "@auth/core/providers/keycloak";
 import type { TokenSet } from "@auth/core/types";
 import { SvelteKitAuth } from "@auth/sveltekit";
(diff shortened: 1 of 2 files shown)


Isak-Kallini (approved):

danieladugyan:
For future reference: I think we ended up concluding that the issue was that `setLanguageTag` was called in a layout server loader. All [load functions are ran concurrently](https://svelte.dev/docs/kit/load#Parallel-loading), so sometimes `setLanguageTag` would get called after the load functions fetching from the database, causing the website to display content in the wrong language.

That still doesn't really make sense to me though - why does it cause the page to switch language? @Fiery-132 @Macludde

Fiery-132:
Well, the documentation isn't exactly great but it seems [the language tag is global](https://github.com/opral/inlang-paraglide-js?tab=readme-ov-file#setting-the-language), and this makes sense considering the observed behaviour. To reproduce the bug on my localhost I had to:

1. Open a tab and switch the language to `en`
2. Open an incognito tab with the language set to `sv` and refresh so the last set tag is `sv`
3. Go back to the original tab and click a link that uses our custom `redirect` (which uses `i18n.resolveRoute()`)
4. After the redirect, the language would be `sv` again

Meaning `setLanguageTag()` had not yet been called and thus the language tag was still `sv` from the last request, since it is global. _However_, the Paraglide-SvelteKit adapter [docs](https://inlang.com/m/dxnzrydw/paraglide-sveltekit-i18n/getting-started) state the opposite:

> On the server `languageTag()` is scoped to the current request, there is no danger of multiple requests interfering. `languageTag()` can safely be called in server-load functions and form actions.

But `i18n.resolveRoute()` uses `languageTag()` and still returns the wrong route so this doesn't seem to be correct.. 🤔 Either way, setting the tag server-side in `hooks` seemed to resolve the issue since it should force the correct tag to be set at the start of every request.

PR #734, opened 2025-02-24
Fix #692 pfp edit button placement

### LINKED ISSUE #692: Unfortunate placement of edit icon for profile picture
### Description

In the recent design change for the profile pages, the button to edit profile pictures was placed at the top of the picture. This is unfortunate as this is where a persons face ends if the picture includes their body. This only happens when the you have access to edit a profile so it is not an issue when viewing the profiles of others unless you have admin rights. 

Image

I would recommend putting the button in the bottom right corner of the profile image since that's where it's usually located. 

### Steps to reproduce

Go to your own profile page to see where the icon is placed.
If you have admin rights, just you will be able to see the icon on all profiles including [mine](https://www.dsek.se/members/ru4786ko-s), where you can see the example above.

### Further information

_No response_

RufusKoggRojder:
Moved the edit button to the bottom right corner of the profile pic and made it circular.

### DIFF
--- a/src/routes/(app)/members/[studentId]/+page@(app).svelte
+++ b/src/routes/(app)/members/[studentId]/+page@(app).svelte
@@ -65,9 +65,9 @@
   {#if canEdit}
     <a
       href="{studentId}/profile-picture"
-      class="btn btn-square glass btn-primary btn-sm absolute left-1/2 top-2 origin-center !-translate-x-1/2"
+      class="btn btn-circle glass btn-primary btn-sm absolute bottom-2 right-0 origin-center !-translate-x-1/2"
     >
-      <span class="i-mdi-edit" />
+      <span class="i-mdi-edit absolute bottom-[0.6rem] right-[0.6rem]" />
     </a>
   {/if}
 </MemberAvatar>


src/routes/(app)/members/[studentId]/+page@(app).svelte
```diff
@@ -65,9 +65,9 @@
   {#if canEdit}
     <a
       href="{studentId}/profile-picture"
-      class="btn btn-square glass btn-primary btn-sm absolute left-1/2 top-2 origin-center !-translate-x-1/2"
+      class="btn btn-circle glass btn-primary btn-sm absolute bottom-2 right-0 origin-center !-translate-x-1/2"
     >
-      <span class="i-mdi-edit" />
+      <span class="i-mdi-edit absolute bottom-[0.6rem] right-[0.6rem]" />
```
Comment (Isak-Kallini) [inline src/routes/(app)/members/[studentId]/+page@(app).svelte:70]: Thanks for fixing this! Some comments/thoughts:
I think we should try to use the built-in "sizes" as much as possible, for example `bottom-2` is the same as `bottom-[0.5rem]`.
Also, can't this be centered?
And lastly, since you're a developer you don't need to work from a fork! You can create branches in the main repo directly :)

Isak-Kallini (approved):

PR #735, opened 2025-02-24
Profile pic fix

oliverlevay:
image
image


JuliusNicklas (approved):
Looks good. Flex is better

Issue #446, opened 2024-08-22
Add a "Register" button


danieladugyan:
There's currently no way for a new user to find our registration page.

Macludde:
If we do this we have to hide it in the app because apple has the rule: "if a user can sign up in your app, they have to be able to delete their account and data". We don't support account deletion, thus can't allow them to sign up in the app.

danieladugyan:
Alright, I suggest we just skip this for now.

PR #737, opened 2025-03-03
Added register button to landing page.

JuliusNicklas:
Pls code review. I might not have done the redirect in the nicest way.
The button is on a the landing page, which is not accessible on mobile. 
#446


Joel-Liden (approved):
From trying it myself, it looks like it brings the user to the correct site, good addition!

Issue #729, opened 2025-02-17
Expense Report Finalization


trilleplay:
### Description

I was originally planning to finish up the fantastic work by @Macludde on the expenses system, but due to other undertakings that take have taken priority, I've not been able to do so.

What must be done from a bookkeeping perspective is to somehow manufacture a record which can be used for bookkeeping purposes, I did manage to do some research in how we transport the record to our bookkeeping system (Fortnox) and they provide a convenient solution for emailing records into one's bookkeeping system.

### Describe the solution you'd like

It would be great if a PDF could be generated and emailed to Fortnox when the "Send to Fortnox/bookkeep" button is pressed, the process for sending the e-mail is just like sending any regular e-mail.

The PDF itself should contain all the data we've collected, one very important request I have is that the images attached to a expense report MUST be provided as seperate pages in the PDF file. So first page in the PDF should be a report of information of when and who reported the expense, who approved the expense, essentially all the data we've collected and display currently. The second page should be the first attachment, the third page be the second attachment and so on...

This PDF should then be a regular attachment to the email that is sent.

The record page (the first page) should be plain/boring, no fancy design, and simple to copy text from.

PR #741, opened 2025-03-04
729 expense report finalization

Macludde:
I'm not gonna lie, I wanted to try out Claude Sonnet 3.7 so most of this code was written in Cursor with that, but honestly I think it did a pretty good job. Of course, I went through the code and fixed some various issues, mostly with typescript and linting, but from an experiment perspective I think it's a success.

This PR implements a button to click "Send to bookkeeping", which generates a PDF as according to #729 and sends an email to a .env-driven recipient address

### DIFF
--- a/.env
+++ b/.env
@@ -79,7 +79,7 @@ PUBLIC_DISABLE_SSR_GLOBALLY=false
 
 
 SMTP_HOST=mailmaster.blossom.dsek.se
-SMTP_PORT=465
+SMTP_PORT=587
 SMTP_USER=_dsekweb_email
 SMTP_PASS=your-smtp-password
 BOOKKEEPING_EMAIL=bookkeeping@example.com 
--- a/.env
+++ b/.env
@@ -82,4 +82,5 @@ SMTP_HOST=mailmaster.blossom.dsek.se
 SMTP_PORT=587
 SMTP_USER=_dsekweb_email
 # SMTP_PASS=your-smtp-password
-BOOKKEEPING_EMAIL=bookkeeping@example.com 
+BOOKKEEPING_EMAIL_TO_ADDRESS=bookkeeping@example.com 
+BOOKKEEPING_EMAIL_FROM_ADDRESS=automatic-expensing@dsek.se
--- a/.env
+++ b/.env
@@ -81,5 +81,5 @@ PUBLIC_DISABLE_SSR_GLOBALLY=false
 SMTP_HOST=mailmaster.blossom.dsek.se
 SMTP_PORT=587
 SMTP_USER=_dsekweb_email
-SMTP_PASS=your-smtp-password
+# SMTP_PASS=your-smtp-password
 BOOKKEEPING_EMAIL=bookkeeping@example.com 
--- a/.env
+++ b/.env
@@ -84,3 +84,4 @@ SMTP_USER=_dsekweb_email
 # SMTP_PASS=your-smtp-password
 BOOKKEEPING_EMAIL_TO_ADDRESS=bookkeeping@example.com 
 BOOKKEEPING_EMAIL_FROM_ADDRESS=automatic-expensing@dsek.se
+BOOKKEEPING_CC_TO_ADDRESS=skattm@dsek.se # comma separated list
--- a/.env
+++ b/.env
@@ -76,3 +76,10 @@ MEILISEARCH_MASTER_KEY=masterKey
 POPPLER_SERVER_URL=http://localhost:8800
 
 PUBLIC_DISABLE_SSR_GLOBALLY=false
+
+
+SMTP_HOST=mailmaster.blossom.dsek.se
+SMTP_PORT=465
+SMTP_USER=_dsekweb_email
+SMTP_PASS=your-smtp-password
+BOOKKEEPING_EMAIL=bookkeeping@example.com 
--- a/src/lib/email/emailService.ts
+++ b/src/lib/email/emailService.ts
@@ -5,6 +5,7 @@ const { SMTP_HOST, SMTP_PORT, SMTP_USER, SMTP_PASS } = env;
 
 const transporter = nodemailer.createTransport({
   host: SMTP_HOST,
+  // port: Number(465),
   port: Number(SMTP_PORT),
   secure: false, // TLS requires secureConnection to be false
   auth: {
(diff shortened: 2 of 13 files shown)


.env
```diff
@@ -75,4 +75,11 @@ MEILISEARCH_MASTER_KEY=masterKey
...
-PUBLIC_DISABLE_SSR_GLOBALLY=false
\ No newline at end of file
+PUBLIC_DISABLE_SSR_GLOBALLY=false
+
+
+SMTP_HOST=mailmaster.blossom.dsek.se
+SMTP_PORT=587
+SMTP_USER=_dsekweb_email
+# SMTP_PASS=your-smtp-password
+BOOKKEEPING_EMAIL=bookkeeping@example.com
```
danieladugyan, line 85: I assumed this would be the address the email would be sent _from_, not to. Maybe I'm just stupid 😄 but a few comments could clarify what these environment variables are used for.

danieladugyan (commented):
Haven't tried it out but looks cool

.env
trilleplay: Fortnox lets us transfer records into our bookkeeping by sending an e-mail to a specially crafted e-mailaddress. So it's correct that we're sending the email _to_ the bookkeeping. The reason it's an environment variable is to make sure it's not easy to send arbitrary junk to our system.

src/lib/email/emailService.ts
```diff
@@ -0,0 +1,52 @@
...
+  port: Number(SMTP_PORT),
+  secure: false, // TLS requires secureConnection to be false
+  auth: {
+    user: SMTP_USER,
+    pass: SMTP_PASS,
+  },
+  tls: {
+    rejectUnauthorized: true,
+    minVersion: "TLSv1.2",
+  },
```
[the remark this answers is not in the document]
trilleplay: Wouldn't surprise me if they're a requirement to be configured this way, email and TLS options are a usually a fight getting right for a given email server.

src/lib/expenses/generatePdf.ts
```diff
@@ -0,0 +1,354 @@
+import { getFullName } from "$lib/utils/client/member";
+import dayjs from "dayjs";
+import { PDFDocument, PDFFont, PDFPage, StandardFonts, rgb } from "pdf-lib";
+import sharp from "sharp";
+import type { ExpandedExpenseForPdf } from "./sendToBookkeeping";
+
+const A4_SIZE: [number, number] = [595.276, 841.89];
```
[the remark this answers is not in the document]
trilleplay: I would assume these are the width and height of a A4 given the variable name, but it likely can be explained better with a comment.

But I'm not sure I agree that a comment is needed.

src/lib/expenses/generatePdf.ts
```diff
@@ -0,0 +1,354 @@
...
+      .rotate()
+      .jpeg()
+      .toBuffer()
+      .then((buffer) => doc.embedJpg(new Uint8Array(buffer))),
+  );
+
+  // Calculate dimensions to fit the image on the page
+  const pageWidth = width - MARGIN * 2;
+  const pageHeight = height - 150;
+  const imageRatio = image.width / image.height;
```
trilleplay, line 278: There is no risk that the transformation of the image may have an affect that could lead to the content being cropped out?

src/lib/expenses/sendToBookkeeping.ts
```diff
@@ -0,0 +1,130 @@
...
+  const expense = await gatherExpenseDataForPdf(prisma, expenseId);
+
+  const pdfBytes = await generateExpensePdf(expense);
+
+  const totalAmount = expense.items.reduce((sum, item) => sum + item.amount, 0);
+
+  if (dev) writePDFToFile(expense, pdfBytes);
+
+  await sendEmail({
+    from: "automatic-expensing@dsek.se",
```
trilleplay, line 104: shouldn't this be an environment variable, my $0.02.

trilleplay: Then again if the provided email credentials lets us send as any dsek user it doesn't matter.

But if the provided credentials lets us send an email as anyone we probably have another bigger problem...

src/lib/expenses/sendToBookkeeping.ts
```diff
@@ -0,0 +1,130 @@
...
+    subject: `Expense Report #${expense.id} - ${getFullName(expense.member)}`,
+    text: `
+  Expense Report #${expense.id}
+  Member: ${getFullName(expense.member)}
+  Date: ${dayjs(expense.date).format("YYYY-MM-DD")}
+  Description: ${expense.description}
+  Total Amount: ${(totalAmount / 100).toFixed(2)} SEK
+  Type: ${expense.isGuildCard ? "Guild Card" : "Private Expense"}
+
+  Please find the detailed expense report attached.
```
trilleplay, line 115: I believe all of this will be null & void. I'm rather certain Fortnox only cares about attachments but I can verify it. 

With that said I wouldn't mind keeping it in and also having it CC'd to `skattm@dsek.se`
I would probably have it setup so it only goes to a seperate inbox but having it would let me search using Gmail, which is a nice to have.

trilleplay (approved):
just comments and wishes on the code. But the flow looks like just what I wished for in my Github issue so I'm very satisfied if this works!

src/lib/email/emailService.ts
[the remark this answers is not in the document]
Macludde: nope, it was just a recommendation by chat

src/lib/expenses/sendToBookkeeping.ts
Macludde: it could be, but it also has to be specified on the site. I'll change it to an env var

src/lib/expenses/sendToBookkeeping.ts
Macludde: Yeah I thought it would be nice to find info at a glance, and be able to search. In case something has gone wrong.

src/lib/expenses/generatePdf.ts
Macludde: I don't think so. Image is scaled to take up entire width, if that causes the height to be larger than the page, it's scaled down to be the maximum height for the page. Image is never distorted or cropped.

src/lib/expenses/generatePdf.ts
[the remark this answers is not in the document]
Macludde: PageSizes existed in pdf-lib

PR #760, opened 2025-04-07
Queue requests in poppler-server

alfredgrip:
Noticed that we failed to sync meeting documents when indexing Meilisearch, probably due to the server being overloaded by the requests sent. This PR ensures that incoming requests are enqueued and processed one by one, which should provide more robustness.


christofergartner (approved):
LGTM :)

PR #761, opened 2025-04-07
Add MinIO health check and view status in admin/debug

### LINKED ISSUE #757: Missing error handling on /home page
### Description

MinIO was temporarily unreachable after an outage. This caused requests to /home to respond with a 500 error. The error log produced by pm2 showed nothing. The only thing indicating an error in the logs was the usual request logging that showed `Status 500`.

I would expect the page to be fine even if MinIO is down and I would expect logs if something goes wrong.

### Steps to reproduce

Presumably this can be reproduced locally by just setting `PUBLIC_MINIO_ENDPOINT` env to an unavailable URL.

### Further information

Same problem on `/documents`

alfredgrip:
closes #757 

Yes, we could just catch the error instead of using `Promise.reject` and skip doing a health check, but this is more idiomatic and I like this way more :)

### DIFF
--- a/src/lib/files/fileHandler.ts
+++ b/src/lib/files/fileHandler.ts
@@ -16,11 +16,31 @@ export type FileData = {
   isDir?: boolean;
 };
 
-const getFilesInFolder = (
+const isMinIOHealthy = async (): Promise<boolean> => {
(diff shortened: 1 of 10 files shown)


danieladugyan (approved):
LGTM!

PR #780, opened 2025-04-24
Fix poppler pipeline

danieladugyan:
Fixes poppler pipeline. Made some other small tweaks while I was at it.

Seems to work: https://github.com/Dsek-LTH/web/actions/runs/14670145722/job/41174546563 (or see OKD)


alfredgrip (approved):
LGTM 💥

PR #781, opened 2025-04-24
Fix price calculation and standardize conversions

danieladugyan:
Multiplying floats caused floating-points arithmetic errors. For example, `19.9 * 100` evaluates to `1989.9999999999998`. Rounding solves that issue.

While I was at it, I updated price calculations to use a dedicated utility function for consistency and accuracy across the application. This change improves maintainability and ensures all price values are converted correctly.

### DIFF
--- a/src/lib/server/shop/tickets/mutations.ts
+++ b/src/lib/server/shop/tickets/mutations.ts
@@ -1,4 +1,5 @@
 import type { TransactionClient } from "$lib/server/shop/types";
+import { convertPriceToCents } from "$lib/utils/convertPrice";
 import type { TicketSchema } from "$lib/utils/shop/types";
 import { PrismaClient, ShoppableType } from "@prisma/client";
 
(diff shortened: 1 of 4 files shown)


Isak-Kallini (approved):
LGTM! Do you know if there was a reason `floor` was used instead of `round`? Or was it just a mistake?

danieladugyan:
> LGTM! Do you know if there was a reason `floor` was used instead of `round`? Or was it just a mistake?

I thought about it for a while but I think it's just a mistake. I couldn't really come up with a reason anyways.

danieladugyan:
I guess this difference is worth pointing out, but I reckon rounding makes more sense[¹](https://slate.com/technology/2019/10/round-floor-software-errors-stock-market-battlefield.html).
```
Math.floor(19.999 * 100) => 1999
Math.round(19.999 * 100) => 2000
```

PR #786, opened 2025-04-25
Bump db dependencies and add missing migration

danieladugyan:
Update Prisma and related dependencies to their latest versions and add a migration to address a known issue with the database schema being reported as out of sync with the migrations folder.


Isak-Kallini:
This does not fix the migration issue? When I run `pnpm migrate` on this branch i get the same migration you added. You can try it out yourself by just running `pnpm migrate` a couple of times, it will add the migration every time

PR #798, opened 2025-05-10
Web load fix and roles list

scol1s0:
Fixed internal error 500, when loading website and added a list for Roles on /admin/debug


danieladugyan (approved):
Nice! ✨

PR #799, opened 2025-05-10
Implement Open Graph Protocol

### LINKED ISSUE #775: Implement Open Graph protocol (to fix link preview)
Open Graph protocol (https://ogp.me/) adds metadata to pages. I believe for our use case, it's best suited for tweaking link previews in e.g. Discord and Facebook.

Basically, it should solve issues like this:
(reported by Hanna)

Image

and instead give a nice custom title, image and description in the preview.

(I'm certain that we had an issue for this a while ago but I can't find it now. I know we have discussed this feature as well, @danieladugyan)

For every page, but probably most relevant for news, events, songs, members, committees, and positions, add Open Graph metadata such as title and image. Perhaps this can be passed in a component like [SetPageTitle.svelte](https://github.com/Dsek-LTH/web/blob/1f260937087b51730e2e02fd58b35ea2599a190b/src/lib/components/nav/SetPageTitle.svelte)

alfredgrip:
Closes #775 

This adds the metadata to many pages. Some pages don't really need SEO, so they are omitted (such as the debug pages).

Unfortunately, I think we need to add this component manually every time. I tried having a default, fallback value in the root `+layout.svelte`, but if I wanted to override it on a specific page, the metadata tags just get duplicated.

### DIFF
--- a/src/lib/components/shop/cart/CartPage.svelte
+++ b/src/lib/components/shop/cart/CartPage.svelte
@@ -2,6 +2,7 @@
   import { invalidate } from "$app/navigation";
   import FoodPreferenceModal from "$lib/components/FoodPreferenceModal.svelte";
   import SetPageTitle from "$lib/components/nav/SetPageTitle.svelte";
+  import SEO from "$lib/seo/SEO.svelte";
   import type { CartLoadData } from "$lib/server/shop/cart/getCart";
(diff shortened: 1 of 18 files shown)


JuliusNicklas (approved):
lgtm
I tested for admin/alerts and app/account and it seems to work.

PR #802, opened 2025-05-10
659 svenska bokstäver för möteshandlingar

Joel-Liden:
For some reason we deliberately remove åäöÅÄÖ from documents before uploading them. I don't know if this is intended behavour or not. 

I could see this making URLs look a bit worse when live, but since it works locally, I don't see why it wouldn't work.

### DIFF
--- a/src/lib/files/utils.ts
+++ b/src/lib/files/utils.ts
@@ -17,13 +17,7 @@ export const prepareNameForFilesystem = (
   fileName: string,
   extension: string | undefined = undefined,
 ) =>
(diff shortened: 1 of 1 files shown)


Joel-Liden:
After some discussion back and forth, we came to the conclusion the change was possibly made for a reason. However, we can not determine why nor test for it. 

If the issue becomes apparent, just go back to the previous version which uses character substitution. (after that version has been tested locally)

DagHemberg (approved):
proceed with caution (aka "im not to blame if this breaks everything 👍️")

PR #813, opened 2025-05-12
fix broken migration

Isak-Kallini:
This pr "fixes" the broken migration by removing the default value of the colum. This seems to be a known bug in prisma, at least I found [this issue](https://github.com/prisma/prisma/discussions/21547) that's very similar (I tried some similar workarounds but couldn't get it to work).

So, the reason default value exists in the first place is because we wanted a default value when the column was created. But now every row in the prod database should have a value, so this should be fine now.


danieladugyan (approved):
spännande tider

PR #815, opened 2025-05-12
fix news creating preview not switching languages

Joel-Liden:
TO TEST FUNCTIONALITY:

_GO TO NEWS_

_CREATE AN ARTICLE_ 
- **S** as the _Swedish_ title with **sssssssssss** as content
- **E** as the _English_ title with **eeeeeeeeee** as content
The preview should show the current content _dynamically_, regardless of language selected in the editor. (identical to event creator)

CHECK THAT THE CREATED NEWS ARTICLE WORKS IN BOTH LANGUAGES

I found that it works on my machine, but it failed previously somehow.

The reason that news previews do not switch language is due to the activeTab (which checks the language of the news to create) is stored in articleForm, whereas the preview is shown in ArticleEditor. 

In events, the file is just eventEditor, and activeTab is used to switch the text shown in the preview.

I simply ported over the same logic. It is a little redundant, but it works. 

It works the same as events now. (changing how previews are shown should not cause problems to the publishing of news as they are separate)


JuliusNicklas (changes requested):
It does not fix the issue

fgren (approved):
LGTM, works for me.

PR #824, opened 2025-05-19
#773 set own phaddergroup

### LINKED ISSUE #773: Users should be able to set their own phaddergroup
### Description

On your profile page you can show which phaddergroup you were a nolla in. However, you can't set it yourself.

It would be nice if you could set it yourself. Maybe your profile page could prompt you to set your phaddergroup?

Currently, a website admin needs to set your phaddergroup for you.

RufusKoggRojder:
Fix #773 by showing a popup on the user page for users who don't have a phadder group, prompting them to select one. The option to skip this is offered but the popup will show up again for when the page is refreshed. The user is prompted to write in the tech support discord if their group is not on the list.


Joel-Liden (approved):
From what I can see, and the little tests I did, this works! 

One thing to consider is that _both the year and program_ can now be altered freely by the user. Something which we might return to if that seems too much.

alfredgrip:
Won't this be very annoying for alumni where we haven't stored the phaddergroups, so there will be none to select? (or only a few are stored, but not the group for the person in question)
I think there should be a button that basically is "don't remind me again", and this answer could be stored in localstorage so we don't have to deal with cookies. Then the user will be reprompted only on new devices

Joel-Liden (changes requested):
The English text has a spelling mistake, introduction is duplicated. 

I initially didn't think about what Alfred said, but that seems like a good idea. (Or maybe a "none/default" option if it's not too troublesome)

PR #826, opened 2025-05-21
Sort songs by creation date

fgren:
Sort songs in the songbook by creation date (latest first) instead of alphabetical sorting, to hopefully counter the problem of having inappropriate songs at the top.


Tomas-Ka (approved):
lgtm

PR #834, opened 2025-06-02
Sync HTML and TXT meeting documents

alfredgrip:
We have support for PDFs, this adds support for HTML and TXT.

I think this is all types of documents that we can sync. I have seen that we also store TEX source code documents, but these are accompanied the compiled PDF.

Also refarctors the syncDocuments file a bit, and tries to use better logging practices.


Isak-Kallini (changes requested):
I tried it out locally with the real minio data, and it throws an error here which crashes the sync: 

Image

PR #870, opened 2025-09-16
Use authentik for managing email aliases

danieladugyan:
### Background
Before this PR, we used to store email aliases, or email groups, in our Postgres database. We had different types of receivers, or members:
- position_ids, i.e `dsek.infu.mastare`
- special receivers, i.e `ordforande@example.se`
- special senders, i.e `ab1234cd-s`

We aggregated all of this information into two public API endpoints: /mail/alias and /mail/alias/senders. Our email server would periodically send a GET request to these endpoints and use the information to update the internal mailing lists/sender lists in Postfix.

We no longer have our own email server. Instead, we use Gmail where email aliases correspond to Google Groups. A member in a Google Group is an email address.

### Changes
This PR therefore replaces all of our previous types of email group members with a single one: an email address, i.e `ordforande@example.se`. This email address can correspond to any of the previous types:
- positions, i.e `dsek.infu.mastare@dsek.se`
- special receivers, i.e `ordforande@example.se` or `ab1234cd-s@user.dsek.se`
- ~~special senders~~ N/A, does not exist in Google Groups.

All email related stuff in the database is deleted – it's all stored directly in authentik instead. This entirely bypasses the issue of having to sync email groups back and forth between Postgres and authentik. Two custom attributes are used in authentik: `mail` for storing the groups email address and `recipients` which stores a list of group members.

### Considerations
There's not really any reason why we'd choose to manage email groups in authentik, since Google is the source of truth anyways. It made more sense in FreeIPA where we could add "normal" groups, such as `dsek.cpu`, as members to email groups. That's not possible in authentik since groups can only have one parent (for `dsek.cpu` the parent is `dsek`).
I think it's fine for the moment, but we might want to consider changing this GUI to just interface directly with Google's API.

### Future work
* ~~We need to figure out a way to migrate all of our old email aliases into authentik. I tried customizing authentik's FreeIPA sync to do that but I couldn't manage it. Instead, we'll need some kind of script that fetches all of our old email groups from Postgres, converts their recipients into email addresses where needed, and adds them to authentik.~~
* It would be nice with some additional validation of the email addresses added as recipients. Obviously we can't really validate that arbitrary email addresses actually exist, but we can at least make sure that email addresses that correspond to groups actually exist in Google.
* There's no way to rename a group using the dsek.se UI (it's trivial to do in authentik).

### DIFF
--- a/src/translations/en.json
+++ b/src/translations/en.json
@@ -40,7 +40,7 @@
   "footer_sourceCode": "Source code",
   "footer_contactAndSourceCode": "Contact and source code",
   "footer_socialMedia": "Social media",
-  "home_info": "<strong>The D-guild within TLTH</strong> is a non-profit organization for students and alumni of the programs <span class=\"text-primary\">Computer Scidnce</span> and <span class=\"text-secondary\">Information and Communication Technology</span> . The guild has social arrangements, business contacts, study supervision, and everything else that helps students and alumni.",
+  "home_info": "<strong>The D-guild within TLTH</strong> is a non-profit organization for students and alumni of the programs <span class=\"text-primary\">Computer Science</span> and <span class=\"text-secondary\">Information and Communication Technology</span> . The guild has social arrangements, business contacts, study supervision, and everything else that helps students and alumni.",
   "home_forCompanies": "For companies",
   "home_nolla": "New student?",
   "home_splash": "Experience the <span class=\"text-primary\">fun</span> of University",
--- a/src/routes/(app)/admin/email-alias/+page.svelte
+++ b/src/routes/(app)/admin/email-alias/+page.svelte
@@ -147,7 +147,7 @@
     <p>
       <!-- eslint-disable-next-line svelte/no-at-html-tags -->
       {@html m.admin_emailalias_removeAliasConfirm({
-        mailaliasName: selectedEmailAlias!.mail,
+        mailaliasName: selectedEmailAlias?.mail ?? "no mail found",
       })}
     </p>
 
--- a/src/routes/(app)/admin/email-alias/[id]/+page.svelte
+++ b/src/routes/(app)/admin/email-alias/[id]/+page.svelte
@@ -37,8 +37,8 @@
   <table class="table table-zebra">
     <thead>
       <tr class="bg-base-200">
-        <th
-          >{m.admin_emailalias_receiver()}
+        <th>{m.admin_emailalias_receiver()}</th>
+        <th>
           <button
             class="btn btn-primary btn-xs float-right px-4"
             onclick={() => addDialog.showModal()}


src/routes/(app)/admin/email-alias/+page.server.ts
```diff
@@ -1,275 +1,60 @@
...
-      type: "success",
-    });
+
+    const { alias, domain } = form.data;
+    const email = `${alias}@${domain}`;
+
+    await createEmailGroup(email);
+    fetchEmailGroups.invalidate();
+
+    redirect(303, `email-alias/${email}`);
```
Isak-Kallini, line 48: This should redirect to `email-alias/[id]`, not email

src/routes/(app)/admin/email-alias/+page.server.ts
```diff
@@ -1,275 +1,60 @@
...
-    return message(form, {
-      message: m.admin_emailalias_addressCreated(),
-      type: "success",
-    });
+
+    const { alias, domain } = form.data;
+    const email = `${alias}@${domain}`;
+
+    await createEmailGroup(email);
+    fetchEmailGroups.invalidate();
```
Isak-Kallini, line 46: I think invalidating the cache is in general nicer to do in the utils methods, so inside createEmailGroup in this case.

src/lib/server/authentik/index.ts
```diff
@@ -22,6 +27,124 @@ function connect(): CoreApi {
...
+  const recipients = new Set(attributes?.["recipients"] ?? []);
+  recipients.delete(email);
+
+  client.coreGroupsPartialUpdate({
+    groupUuid: id,
+    patchedGroupRequest: {
+      attributes: { ...attributes, recipients: [...recipients] },
+    },
+  });
+}
```
Isak-Kallini, line 89: email groups cache should be invalidated in these two functions as well

Isak-Kallini (changes requested):
Also, there's a lot of text that isn't translated

src/routes/(app)/admin/email-alias/+page.server.ts
```diff
@@ -1,275 +1,60 @@
...
-        },
-      },
-    });
-    return message(form, {
-      message: m.admin_emailalias_addressCreated(),
-      type: "success",
-    });
-  },
+  create: async ({ locals, request }) => {
+    authorize(apiNames.EMAIL_ALIAS.CREATE, locals.user);
```
Tomas-Ka: Creating a new Email alias (which I assume is here) with the same name as an already existing alias will create an unclear error about an authentik error code with no additional information. Nothing is output in the terminal either.

Tomas-Ka: this is fixed in 0cfe8ea

Tomas-Ka:
the latest commit e3ba7af has some potentially hacky solutions to get past the type-checker, which I'd like some further input on. Apart from that and the other comments that are still open in this thread this is up to date with main and ready to merge!

Tomas-Ka (approved):

src/translations/en.json
```diff
@@ -40,7 +40,7 @@
   "footer_sourceCode": "Source code",
   "footer_contactAndSourceCode": "Contact and source code",
   "footer_socialMedia": "Social media",
-  "home_info": "<strong>The D-guild within TLTH</strong> is a non-profit organization for students and alumni of the programs <span class=\"text-primary\">Computer Science</span> and <span class=\"text-secondary\">Information and Communication Technology</span> . The guild has social arrangements, business contacts, study supervision, and everything else that helps students and alumni.",
+  "home_info": "<strong>The D-guild within TLTH</strong> is a non-profit organization for students and alumni of the programs <span class=\"text-primary\">Computer Scidnce</span> and <span class=\"text-secondary\">Information and Communication Technology</span> . The guild has social arrangements, business contacts, study supervision, and everything else that helps students and alumni.",
```
Isak-Kallini: Added a spelling error??

PR #889, opened 2025-10-06
exit early and silence errors from docker rm

lego-eden:


ManneG (approved):
Stuff i forgot before, but it works with this. :+1:

Issue #716, opened 2025-02-10
QR scan page


ScottTolleback1:
### Description

Scan and get status of ticket, then maybe redirect.

PR #895, opened 2025-10-06
Consume endpoint

ekvanox:
Adds the ability to scan tickets via qr codes, and validating them server side.

Users get a qr code:
*URL: /shop/inventory/{ticket_id}*

image

Admin can consume the ticket:
*URL: /events/{event_slug}/scan/{ticket_id}*
image

New QR Code option under the admin tab:
image

Closes https://github.com/Dsek-LTH/web/issues/716

### DIFF
--- a/src/lib/server/shop/consumable.ts
+++ b/src/lib/server/shop/consumable.ts
@@ -1,7 +1,7 @@
-import type { PrismaClient } from "@prisma/client";
+import type { ExtendedPrisma } from "$lib/server/extendedPrisma";
 
 export const consumeConsumable = async (
-  prisma: PrismaClient,
+  prisma: ExtendedPrisma,
   consumableId: string,
 ): Promise<Message> => {
   try {


src/lib/components/QRCodeScanner.svelte
```diff
@@ -0,0 +1,59 @@
...
+  import { useQRScanner } from "$lib/hooks/useQRScanner";
+
+  export let onScan: (text: string) => void;
+
+  let videoElement: HTMLVideoElement;
+  let errorMessage = "";
+
+  const { initialize } = useQRScanner();
+
+  $: if (videoElement) {
```
fgren: Since we are slowly migrating to Svelte 5, it would be nice to have this in Svelte 5 syntax. (however, I think it's fine to merge anyway)

src/lib/components/QRCodeScanner.svelte
```diff
@@ -0,0 +1,59 @@
...
+  #qr-reader {
+    width: 100%;
+  }
+
+  .error-message {
+    color: red;
+    text-align: center;
+    padding: 1rem;
+  }
+</style>
```
fgren: For consistency, this should be using Tailwind instead of plain CSS.

src/routes/(app)/events/[slug]/scan/[consumable]/+page.svelte
```diff
@@ -0,0 +1,153 @@
+<script lang="ts">
+  import { page } from "$app/stores";
+  import dayjs from "dayjs";
+  import { invalidateAll } from "$app/navigation";
+
+  const slug = $page.params["slug"];
+  export let data;
+
+  $: consumable = data.consumable;
```
Comment (fgren) [inline src/routes/(app)/events/[slug]/scan/[consumable]/+page.svelte]: Same as before about Svelte 5.

PR #900, opened 2025-10-06
Start extracting mandate and mandate editing from positions page

### LINKED ISSUE #571: Add an admin interface for managing positions
### Description

Whenever positions change due to changes in our governing documents, we have to manually ensure these changes are reflected in our application. For instance, to add a new position we have to connect directly to the database and add a new row in the `Positions` table.

Additionally, these positions are assigned an id which needs to correspond to a group in Keycloak. If a new position is created in our application database, a corresponding new group needs to be created in Keycloak.

Add an admin interface somewhere over on `/committees/:slug` to make it possible for normal users to manage positions in a committee, instead of just DWWW-developers with database access.

The current solution of connecting to the database is workable, but changes happen more often than you'd think and it's somewhat time consuming to make changes directly to the database.

ManneG:
Closes #571

### DIFF
--- a/src/routes/(app)/positions/[id]/+page.svelte
+++ b/src/routes/(app)/positions/[id]/+page.svelte
@@ -1,19 +1,15 @@
 <script lang="ts">
   import { page } from "$app/stores";
-  import ClassBadge from "$lib/components/ClassBadge.svelte";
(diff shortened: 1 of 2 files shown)


RufusKoggRojder (approved):
LGTM :+1:

PR #901, opened 2025-10-06
It can now store files at minio sandbox

JuliusNicklas:
One must write the user and password for minio first, though


Tomas-Ka (approved):
maybe some superfluous comments, but lgtm

PR #771, opened 2025-04-14
Fixed languages (closes #688)

Fiery-132:
Changed database so fields `name`, `title`, etc. now are `nameSv` and `titleSv`. This allows reading both english and swedish version regardless of which language the website is in. The website now uses a prisma extension to dynamically insert the `name` or `title` field for when the current language version is desired. Also added typing to add these fields to the prisma client.

## Todo:

- [ ] Change instances of `model.name` to `model.nameSv` where both languages should be displayed
- [ ] Change functions to use the new `ExtendedPrisma` type instead of `PrismaClient`
- [ ] Change occurences of `modelName` types from prisma to `ExtendedPrismaModel[modelName]` instead

PR #905, opened 2025-10-07
Fix alerts form and editing recurring events

Isak-Kallini:
I found two problems left from #771 
- Create a recurring event and try to edit it
- Try to create an alert
(you can test this on staging.dsek.se if you want to confirm)


RufusKoggRojder (approved):
Looks good :+1:

PR #914, opened 2025-10-17
install pnpm via npm in poppler dockerfile

Fiery-132:
Newest node versions no longer bundle corepack so this build broke 😢


Isak-Kallini:
I think we should stick with 24 since 25 will be EoL relatively soon

Issue #284, opened 2024-04-22
:zap: Photo archive


fnurkla:
```[tasklist]
### Tasks
- [ ] View single images
- [ ] CRUD images
- [ ] Browse image albums
- [ ] View image metadata
- [ ] Tag people à la `old.dsek.se`
- [ ] (Set image/tag as profile picture)
- [ ] Report image — if you don't like an image of yourself or the like
- [ ] Comments?
- [ ] Connect images to events?
- [ ] Allow users to submit images
```
```[tasklist]
### MinIO
- [ ] Access control
- [ ] Minio bucket for storage
- [ ] Filestash for management
```

fnurkla:
After talking a bit with Oskar, we said that a Minio bucket will work well for our purposes. Image metadata will be in the database though.

trettiofem:
Mockup: https://drive.google.com/drive/folders/1KZlu9F5ydpp2smTtDynj2tg0aXu7qVHJ?usp=drive_link

PR #921, opened 2025-10-20
Add Image archive

fgren:
Adds a gallery where images can be uploaded and browsed. All files and data are stored in Minio, with the folder structure defining the albums.

This will be superseded by the full archive, when it is built. In the meantime, it is intended as a complement to the Facebook gallery.

Partially (and temporarily) implements #284

### DIFF
--- a/src/translations/en.json
+++ b/src/translations/en.json
@@ -844,7 +844,7 @@
   "gallery_upload_files": "Files",
   "gallery_upload_upload": "Upload",
   "gallery_photographer": "Photographer",
-  "gallery_editor": "Editor"
+  "gallery_editor": "Editor",
   "positions_removeMandate": "Remove mandate",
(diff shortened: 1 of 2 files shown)


src/routes/(app)/gallery/upload/+page.server.ts
```diff
@@ -0,0 +1,40 @@
...
+          message: e instanceof Error ? e.message : `${e}`,
+          type: "error",
+        },
+        { status: 500 },
+      );
+    }
+    // Clear file and name
+    form.data.files = null as unknown as File[]; // will work, but not type correct
+    form.data.name = "";
+    redirect(303, "/gallery/album/" + form.data.date + " " + form.data.name);
```
Isak-Kallini: Setting files and name to empty is not necessary since we redirect anyways?

PR #927, opened 2025-10-27
Add position based access admin interface

Isak-Kallini:
With the current admin interface `/admin/access` you have to click into each apiname to see who has access to that. I've added a page where it's displayed per position instead, and instead of going to a separate page to view a position i used a collapse.

image


src/routes/(app)/admin/access/positions/+page.server.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export const load: PageServerLoad = async ({ locals }) => {
+  const { prisma } = locals;
+  const accesspolicies = await prisma.accessPolicy.findMany({
+    select: { role: true, apiName: true, id: true },
+  });
+  const posToAccessPolicies = new Map<
+    string,
+    Array<{ apiName: string; id: string }>
+  >();
```
danieladugyan: This approach could cause some confusion for users. For each "position" it only lists explicit access policies, not inherited ones. That's quite obvious when you look at the code, but it might not be in line with what you'd expect.

Solving that makes this a lot messier so IDK if it's worth fixing. For instance, even if you fix inherited access policies there's also the issue of showing positions that don't currently have any explicit access policies (but maybe you'd like to assign them a new one from this new view).

PR #939, opened 2025-11-03
Fix election dates

RufusKoggRojder:
For real this time (maybe). This is done by removing the client-side date localisation. Just like in #935 it cant really be tested in localhost and has to be tested in the staging environment.


fgren (approved):
good?

PR #966, opened 2025-11-10
add button

Isak-Kallini:
Button was already added since I was testing with it on the new-design branch. I updated it to the latest version and added correct colours


src/lib/components/ui/button/button.svelte
```diff
@@ -86,31 +76,8 @@
...
-              // TL;DR: Under 100ms delay, the user feels it as instant. Over 1000ms delay, the user feels it as a separate task. 500ms is a sweet spot to show a loading spinner to tell the user something is happening.
-            }, 500);
-            await onclick(e); // svelte doesn't recoginize that this is non-nullable
-            clearTimeout(timeout);
-            isLoading = false;
-          } else {
-            await onclick(e); // svelte doesn't recoginize that this is non-nullable
-          }
-        }
-      : undefined}
```
Isak-Kallini: All the loader-stuff was removed because the library removed it in the latest version

Issue #928, opened 2025-10-27
Create a scheduling service


Isak-Kallini:
Currently, we don't really have a good way to schedule something to run at a later point in time. The naive solution is to use `setTimeout`, but this will be unreliable since the timeout could disappear if the pod it is running in restarts. So we want something tha cant:
- Handle restarts
- Guarantee something is only run once

A use case for this is scheduling article publication (#690). It is simple to add a `publishedAt` field and just use a filter to show it to the user after the publish time. This misses one very important thing though, notifications. This is something that has to be run when the article is published.

### Design
This service should be as simple and generic as possible. We should have an enpoint, something like `POST /schedule` which we call when we want to schedule something. The supplied data should be: time to run, endpoint to call, body (id of article to send notification for example).

The service should schedule a call to the enpoint with the supplied body. It should also store this information in a database along with a `hasExecuted` field. This is so it can handle restarts, when it starts up it should read all unexecuted jobs and schedule them. If there's something that should have been executed with a false `hasExecuted` this should be executed immediately and then marked as executed.

The endpoint in web that is called should usually not be able to be called by anyone. Web and the server should use a shared secret for authentication. Look at the sync cronjob for an example.

Issue #690, opened 2025-01-21
Schedule announcements


christofergartner:
### Description

When posting in "News", it is not currently possible to schedule a posting. This would be quite convenient for when you want to write an announcement, but want it to drop at a specific point in time.

### Describe the solution you'd like

Add a tickbox to allow scheduling a news post for later, with the user inputting time.

### Describe alternatives you've considered

None.

PR #977, opened 2025-11-12
Scheduler service

Fiery-132:
Implements #928 and #690

This PR introduces a new scheduler service exposed as a REST API written in Go. The service supports creating, fetching, updating, and deleting scheduled tasks via POST, GET, PATCH, and DELETE HTTP requests.

The scheduler is used to add functionality for scheduled publishing of news articles. When an article is set to publish at a future date with notifications enabled, it is registered with the scheduler, which triggers notifications at the specified date and time. Scheduled articles can also be updated after creation.

News creation form with new field:
image

Scheduled articles can be seen on the news page:
image

---

### TODO:
- [x] add dockerfile
- [x] support editing of scheduled articles

### DIFF
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -71,4 +71,3 @@ jobs:
           context: "{{defaultContext}}:scheduler-service"
           push: true
           tags: registry.dsek.se/web-staging/scheduler:${{ github.ref_name }},registry.dsek.se/web-staging/scheduler:latest
-
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -50,7 +50,6 @@ jobs:
           push: true
           tags: registry.dsek.se/web/poppler:${{ github.ref_name }},registry.dsek.se/web/poppler:latest
 
-
   scheduler:
     runs-on: ubuntu-latest
     steps:
@@ -69,4 +68,3 @@ jobs:
           context: "{{defaultContext}}:scheduler-service"
           push: true
           tags: registry.dsek.se/web/scheduler:${{ github.ref_name }},registry.dsek.se/web/scheduler:latest
-


scheduler-service/main.go
```diff
@@ -0,0 +1,56 @@
...
+
+	if err := openDatabaseConnection(&db); err != nil {
+		log.Fatal("Failed to connect to database:", err)
+	}
+
+	if err := db.AutoMigrate(&ScheduledTask{}); err != nil {
+		log.Fatal("Failed to migrate database:", err)
+	}
+
+	if scheduledTasks, err := gorm.G[ScheduledTask](
```
davidagardh: Break out assignment into its own statement

scheduler-service/taskHandler.go
```diff
@@ -0,0 +1,146 @@
...
+	HasExecuted  bool
+	CreatedBy    *string
+}
+
+type CreateScheduledTaskResponse struct {
+	ScheduledTaskID uint `json:"scheduledTaskID"`
+}
+
+var (
+	taskTimers   = make(map[uint]*time.Timer)
```
davidagardh: Group these two in a struct to communicate they are connected

scheduler-service/databaseHandler.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"fmt"
+	"os"
+
+	"gorm.io/driver/postgres"
+	"gorm.io/gorm"
+)
+
+func openDatabaseConnection(db **gorm.DB) error {
+	host, user, password, name, port := os.Getenv(
```
davidagardh: Nit: don't like this line breaking. Maybe inline these variables in the `dsn` assignment

scheduler-service/rateLimitMiddleware.go
```diff
@@ -0,0 +1,69 @@
...
+	"net"
+	"net/http"
+	"sync"
+	"time"
+
+	"golang.org/x/time/rate"
+)
+
+var (
+	limiters = make(map[string]*trackedLimiter)
```
davidagardh: Group in struct

src/lib/news/server/actions.ts
```diff
@@ -280,13 +124,47 @@ export const createArticle: Action = async (event) => {
...
+      { ...result, tags, notificationText },
+      publishTime,
+      form,
+      m.news_errors_schedulingFailed(),
+      m.news_articleScheduled(),
+      "/news",
+      event,
+    );
+
+    if (isActionFailure(scheduledResult)) {
```
davidagardh: You're even writing `if err != nil { return err }` in your typescipt code! Rob Pike would be proud

scheduler-service/authMiddleware.go
```diff
@@ -0,0 +1,110 @@
...
+		if err != nil {
+			log.Printf("Failed to parse JWT: %s", err)
+			http.Error(w, "Unauthorised", http.StatusUnauthorized)
+
+			return
+		}
+
+		subject, _ := token.Subject()
+		if subject == "" {
+			log.Printf("Invalid subject in JWT: %s", err)
```
davidagardh: Using err which is guaranteed to be nil

scheduler-service/main.go
```diff
@@ -0,0 +1,56 @@
...
+		db,
+	).Where("has_executed = ?", false).
+		Find(context.Background()); err != nil {
+		log.Println("Error fetching scheduled tasks:", err)
```
davidagardh: Question: Is it fine to continue if we can't fetch scheduled task or should this also be a fatal event?

davidagardh (changes requested):

scheduler-service/main.go
Fiery-132: Nice catch :+1:. No, that's probably not fine since it would ignore all old, unexecuted, tasks until next restart. I have fixed this in the latest revision.

src/routes/(app)/news/+page.server.ts
```diff
@@ -27,11 +27,20 @@ export const load: PageServerLoad = async ({ locals, url }) => {
     }),
     getAllTags(prisma),
   ]);
+  const scheduledArticles = await prisma.article.findMany({
```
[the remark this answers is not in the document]
Fiery-132: They should not haha - I missed a filter there. Should be fixed now.

PR #1005, opened 2025-11-17
New design: Add admin doors management interface

danieladugyan:
Introduce a new layout for managing doors in the admin section, including a form for door access rules, an empty state page, and a dialog for deleting access rules.

image

### To-do
- [x] Show whether rule is allow or ban
- [x] Translations
- [x] Drop commits included in other PRs (~~#1001 #1006~~)

### Future work
- [x] Require reason and end date for member rules
- [x] Require reason for ban rules
- [x] Don't allow banning groups since this currently has no effect (or implement group banning)
- Give user feedback when submitting form (success or errors)


Isak-Kallini (changes requested):
- [x] Loop for doors in layout.svelte should use door.name not door.id
- [x] Should not default end date
- [x] When giving access to member it should not say valfritt for required fields
- [ ] Should we only allow stilid in db? People without accounts?

Isak-Kallini (approved):
Maybe add some feedback when rule is created?

PR #1013, opened 2025-11-19
Possible fix for language switcher

Fiery-132:
Currently there are issues with displaying the website in any language other than the user's set preference. This PR fixes that by setting a session cookie that overrides the user's settings. There are other possible solutions (like updating the db instead of a cookie, for example) but this was the best I could come up with.

### DIFF
--- a/src/hooks.server.ts
+++ b/src/hooks.server.ts
@@ -8,6 +8,7 @@ import {
   isAvailableLanguageTag,
   setLanguageTag,
   sourceLanguageTag,
+  type AvailableLanguageTag,
 } from "$paraglide/runtime";
 import Authentik, {
   type AuthentikProfile,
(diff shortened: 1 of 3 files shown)


Fiery-132:
I considered that but decided against it for two main reasons.

1. The feature is mainly for international students who don't speak Swedish. If they still switch to Swedish, it's likely temporary, and they wouldn't expect that choice to persist in their settings. That's why I felt a session cookie worked well, since it should get cleared on e.g. app restarts.
2. It's not a huge issue, but persisting the choice could make it easier to spam the db with unnecessary requests. It would also introduce a _slightly_ longer delay when switching languages.

However, I also think that this setting is a bit hidden, so I do see a benefit in making it easier to toggle. A good start might be adding it to the onboarding page? We could also add a pop-up for anyone that hasn't selected a language, like we do with food pref? Or allow the switcher to update the db, but then remove the option from settings?

PR #1015, opened 2025-11-22
Add settings page

Isak-Kallini:
Waiting for switch component

### DIFF
--- a/src/routes/(app)/+layout.svelte
+++ b/src/routes/(app)/+layout.svelte
@@ -1,9 +1,9 @@
-<script>
+<script lang="ts">
   import Toast from "../Toast.svelte";
   import Alert from "$lib/components/Alert.svelte";
-  import { languageTag } from "$paraglide/runtime";
   import Footer from "./Footer.svelte";
   import Navbar from "./Navbar.svelte";
+  import { getLocale } from "$paraglide/runtime";
 
   const { data, children } = $props();
 </script>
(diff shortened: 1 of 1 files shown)


src/routes/(app)/settings/+page.svelte
```diff
@@ -1,5 +1,199 @@
...
+    </div>
+    <div class="ml-10">
+      <h4>{m.setting_mode()}</h4>
+      <Select.Root
+        type="single"
+        name="mode"
+        onValueChange={(value) => setMode(value as "dark" | "light" | "system")}
+      >
+        <Select.Trigger class="w-36"
+          >{modeTranslation(data.theme)}</Select.Trigger
```
ManneG: This only says "Dark" for me, even when selecting system or light (it changes the page successfully though)

image

src/routes/(app)/settings/+page.svelte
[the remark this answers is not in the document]
ManneG: One proposal for how it could work is, add a third "language option" being "no preference". This would be the default. If you have selected a preference it would (using a hook) always redirect you to the part of the site with you language of choice. This would then also hide the language button in the header as that would be redundant and do nothing.

Selecting english, and then going to `/news` would then redirect you to `/en/news` unless you went back to settings and selected no preference

Isak-Kallini:
Addressed comments. The language switching was broken on the settings page because of the select language dropdown and I didn't manage to fix it so I just commented it out for now so we can get the pr merged

biljett-bugg.dump
Fiery-132: pls remove

PR #1026, opened 2025-12-01
Choicebox

JuliusNicklas:
Added choicebox and choiceradiobox components. Feedback on naming is welcome :)


Joel-Liden (approved):
lgtm! (if the tests pass) Potentially the files should be placed in a different folder but that can be done easily later.

PR #1030, opened 2025-12-07
New design: Add News pages

fgren:
Includes:
- News overview page, with pagination and search (/news)
- Article page (/news/[slug])
- Article creation
- Article editing

### DIFF
--- a/src/routes/(app)/Navbar.svelte
+++ b/src/routes/(app)/Navbar.svelte
@@ -184,7 +184,7 @@
               aria-label="menubar"
               size="icon-lg"
               variant="ghost"
-              class="md-nav:hidden block p-1.5"><X /></Button
+              class="md-nav:hidden inline-flex p-1.5"><X /></Button
             ></Drawer.Close
           >
         {:else}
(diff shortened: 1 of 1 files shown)


src/routes/(app)/news/create/+page.svelte
ManneG: Going to this route when logged out gives status 500 when it should give 401. Change `+page.server.ts` to throw 401 before line 17 if user is not logged in

src/routes/(app)/news/+page.svelte
```diff
@@ -18,17 +21,17 @@
   }}
 />
 
+<div class="flex flex-row gap-2 py-4">
+  <div class="flex-1 gap-2 md:flex-row md:items-end">
+    <NewsSearch />
+  </div>
+  <a href="/news/create"><Button>+ {m.news_create()}</Button></a>
```
ManneG: Do not display this button when logged out (see my comment on `src/routes/(app)/news/create/+page.svelte`)

src/routes/(app)/news/+page.svelte
```diff
@@ -18,17 +21,17 @@
...
+</div>
 <div class="space-y-4">
```
ManneG: Before the article grid, add pagination here as well, as the content is tall enough to justify it

image

src/routes/(app)/news/ArticleForm.svelte
```diff
@@ -0,0 +1,192 @@
...
+    dt.items.add(f);
+    filelist = dt.files;
+  });
+  function onFileUpload() {
+    // @ts-expect-error -- expected
+    $form.images = [...filelist];
+  }
+</script>
+
+<form
```
ManneG: I have issues submitting the form. Don't know why

src/routes/(app)/news/create/+page.svelte
ManneG: Ok. Improves usability in dev if it responds with 401 since that would not bloat the console log, but works fine as is i guess

PR #1038, opened 2025-12-23
New design: Add onboarding page

fgren:
image

### DIFF
--- a/src/routes/(app)/onboarding/+page.svelte
+++ b/src/routes/(app)/onboarding/+page.svelte
@@ -32,7 +32,6 @@
       data.member &&
       data.member.firstName &&
       data.member.lastName &&
-      data.member.email &&
       data.member.classProgramme &&
       data.member.classYear
     ) {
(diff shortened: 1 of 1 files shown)


src/routes/(app)/onboarding/+page.svelte
```diff
@@ -1,5 +1,195 @@
...
+  import { page } from "$app/state";
+  import Languages from "@lucide/svelte/icons/languages";
+
+  let { data }: { data: PageData } = $props();
+  const { form, errors, constraints, enhance } = superForm<UpdateSchema>(
+    data.form,
+  );
+  onMount(() => {
+    if (
+      data.member &&
```
ManneG: This does not work for me. I stay on the onboarding page instead of getting redirected.

### Steps to reproduce

1. Run `pnpm seed` to reset user
2. Start server and go to `/onboarding`
3. Click `Save`
4. Go to `/onboarding`

fgren: This is caused by the email field not getting set on the user. Removing that check for now.

ManneG (approved):
Not setting email is weird, but hopefully that is fine

PR #1040, opened 2025-12-31
Remove authentik sync code 💣

danieladugyan:
This removes all code related to syncing with authentik — mandates, emails and email aliases. See this article for motivation.

Note the email sync is actually not implemented yet in n8n, but it's not urgent and can easily be added. Mandate sync should really be removed before this evening though.


Isak-Kallini (approved):
Remove email-alias from navbar as well

danieladugyan:
n8n now syncs both email addresses and names from database -> authentik

PR #1056, opened 2026-01-21
fix committee icon fallbacks

fgren:


Fiery-132 (approved):
copilot said it's ok 👍

PR #1135, opened 2026-03-11
Cafe page typing fix

Tomas-Ka:
### 🧩 Summary

Fixes the issues had in #1077.

### 🔗 Related issues (if any)

Closes nothing because we never made an issue

before: 500 server error page (was gone in production because of hotfix)
now: successfully renders webpage

### 💬 Other information

There is a way to generate the typing of the enum programatically with a prisma generate script / macro or something similar, but afaik that's not something we've done previously, and frankly I was dissuaded because it seemed complicated. This was a good solution to me.

### DIFF
--- a/src/routes/(app)/committees/cafe/types.ts
+++ b/src/routes/(app)/committees/cafe/types.ts
@@ -11,7 +11,6 @@ export const TimeSlot = {
 } as const satisfies Record<string, PrismaTimeSlot>;
 export type TimeSlot = (typeof TimeSlot)[keyof typeof TimeSlot];
 
-
 export const scheduleForm = z.object({
   date: z.date(),
   worker: z.string().optional(),


src/routes/(app)/committees/cafe/+page.server.ts
```diff
@@ -4,7 +4,7 @@ import { isAuthorized } from "$lib/utils/authorization";
 import { committeeActions, committeeLoad } from "../committee.server";
 import * as m from "$paraglide/messages";
 import { error, fail } from "@sveltejs/kit";
-import { TimeSlot } from "@prisma/client";
+import { TimeSlot } from "./types";
```
Isak-Kallini: What if this row is `import type { TimeSlot } from "@prisma/client"; `? Can't test right now but this could maybe fix the problem

Tomas-Ka: (which means that importing only the type unfortunately won't solve my problem)

PR #1169, opened 2026-04-27
fix navbar

### LINKED ISSUE #1162: New design: fix navbar content
The current navbar content is a bit of a mess, we need to settle on how it should look like. I propose that we keep it the same or at least very similar to the old design so that it's not too jarring on beta.dsek.se when you switch between the two designs while we transition

lego-eden:
### 🔗 Related issues (if any)

Closes #1162

### DIFF
--- a/src/routes/routes.ts
+++ b/src/routes/routes.ts
@@ -1,6 +1,5 @@
 import * as m from "$paraglide/messages";
 
-import nollningImage from "./(nollning)/nollning/(photos)/staben25_9x16.webp";
 // bottom-nav: Show in the bottom navigation bar
 // home-link: Show on the home page
 // none: Don't show anywhere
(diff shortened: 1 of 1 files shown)


src/routes/routes.ts
```diff
@@ -25,111 +24,167 @@ export type Route = {
...
+          path: "/documents/governing",
         },
         {
-          title: m.nav_nollning(),
-          description: m.nav_nollning_description(),
+          title: m.documents_meetingDocuments(),
           accessRequired: null,
           appBehaviour: "none",
-          path: "/nollning",
+          path: "/documents/meeting",
```
fgren: this should be `/documents`

src/routes/routes.ts
```diff
@@ -25,111 +24,167 @@ export type Route = {
...
+          path: "/board",
         },
         {
-          title: m.nav_board(),
-          description: m.nav_board_desc(),
+          title: m.nav_committees(),
           accessRequired: null,
           appBehaviour: "none",
-          path: "/board",
+          path: "/committees",
```
fgren: this should be /about

fgren (changes requested):
two links are broken, otherwise good!

fgren (approved):

PR #1174, opened 2026-04-28
Fix shell scripts according to shellcheck and shfmt

SLUCHABLUB:
### 🧩 Summary

Fixes all shell scripts in the repo.
- I switched to the recommended shebang for bash
- I fixed `shellcheck` warnings
- I ran `shfmt`


Tomas-Ka (approved):
looks good to me!

PR #1180, opened 2026-05-06
sort SRD meetings correctly

fgren:
### 🧩 Summary
Sorts the new SRD meeting format correctly:

Before: SRD1, SRD10, SRD8, SRD9
After: SRD1, SRD8, SRD9, SRD10

### DIFF
--- a/src/routes/(app)/documents/+page.svelte
+++ b/src/routes/(app)/documents/+page.svelte
@@ -34,11 +34,26 @@
       value: "other",
     },
   ];
-  $: meetings = Object.keys(data.meetings).sort((a, b) =>
+  $: meetings = Object.keys(data.meetings).sort((a, b) => {
+    if (type === "board-meeting") {
+      return b.localeCompare(a, "sv");
+    } else if (type === "SRD-meeting" && a.startsWith("SRD")) {
+      return (
+        // Current format
+        Number.parseInt(b.split("SRD")[1] ?? "0") -
+        Number.parseInt(a.split("SRD")[1] ?? "0")
+      );
+    } else if (type === "SRD-meeting") {
+      return ("T" + a).localeCompare(b, "sv"); // Sort other SRD meetings below current format
+    } else {
+      return a.localeCompare(b, "sv");
+    }
+  });
+  /*
     type === "board-meeting" || type === "SRD-meeting"
       ? b.localeCompare(a, "sv")
       : a.localeCompare(b, "sv"),
-  );
+  );*/
   $: canCreate = isAuthorized(
     apiNames.FILES.BUCKET(PUBLIC_BUCKETS_DOCUMENTS).CREATE,
     data.user,
--- a/src/routes/(app)/documents/+page.svelte
+++ b/src/routes/(app)/documents/+page.svelte
@@ -49,11 +49,6 @@
       return a.localeCompare(b, "sv");
     }
   });
-  /*
-    type === "board-meeting" || type === "SRD-meeting"
-      ? b.localeCompare(a, "sv")
-      : a.localeCompare(b, "sv"),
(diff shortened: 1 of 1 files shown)


src/routes/(app)/documents/+page.svelte
```diff
@@ -34,11 +34,26 @@
...
+        Number.parseInt(b.split("SRD")[1] ?? "0") -
+        Number.parseInt(a.split("SRD")[1] ?? "0")
+      );
+    } else if (type === "SRD-meeting") {
+      return ("T" + a).localeCompare(b, "sv"); // Sort other SRD meetings below current format
+    } else {
+      return a.localeCompare(b, "sv");
+    }
+  });
+  /*
```
Tomas-Ka: could we not remove this comment?

src/routes/(app)/documents/+page.svelte
```diff
@@ -34,11 +34,26 @@
       value: "other",
     },
   ];
-  $: meetings = Object.keys(data.meetings).sort((a, b) =>
+  $: meetings = Object.keys(data.meetings).sort((a, b) => {
+    if (type === "board-meeting") {
```
Tomas-Ka, line 38: I don't like that we're using static strings here instead of an enum or something. How hard would it be to rework this to use an enum somewhere to get lsp support?

src/routes/(app)/documents/+page.svelte
```diff
@@ -34,11 +34,26 @@
...
+      return b.localeCompare(a, "sv");
+    } else if (type === "SRD-meeting" && a.startsWith("SRD")) {
```
Tomas-Ka, line 40: should probably be a.lower().startsWith("srd")

src/routes/(app)/documents/+page.svelte
```diff
@@ -34,11 +34,26 @@
...
+      return (
+        // Current format
+        Number.parseInt(b.split("SRD")[1] ?? "0") -
```
Tomas-Ka, line 43: Should probably have some form of error handling?

src/routes/(app)/documents/+page.svelte
```diff
@@ -34,11 +34,26 @@
...
+        Number.parseInt(a.split("SRD")[1] ?? "0")
+      );
+    } else if (type === "SRD-meeting") {
+      return ("T" + a).localeCompare(b, "sv"); // Sort other SRD meetings below current format
```
Tomas-Ka, line 47: Should probably add a comment for why this is needed, say
// This is required due to the format used in 2026

Tomas-Ka (changes requested):
Would like you to just check over some things before merging, but it overall looks good and these are mostly nitpicks.

src/routes/(app)/documents/+page.svelte
fgren: yes

src/routes/(app)/documents/+page.svelte
fgren: it would be some work, out of scope

src/routes/(app)/documents/+page.svelte
fgren: unneeded, the upload page tells you to only name it SRD with capital letters

src/routes/(app)/documents/+page.svelte
fgren: i do not think it is needed

src/routes/(app)/documents/+page.svelte
Tomas-Ka: will make a new PR for this then

Tomas-Ka (approved):
looks good!

PR #1189, opened 2026-05-14
New devcontainers onboarding

### LINKED ISSUE #566: It should be easier to understand if the database is off
### Description

New developers often understand enough to run `pnpm dev`, but they might not realize that they have to start the database first. Thus, they often start the development server without the db running and when they do the error message is not great. It says "can't reach the database" but it's obscured by a large stack trace.

### LINKED ISSUE #570: Document how to start the database
### Description
We currently only have documentation on how to setup the database initially. Add documentation about how to start the database once it's setup. This should be pretty obvious to anyone using Docker Desktop, but it's especially worth clarifying for users of the Docker CLI.

davidagardh:
### 🧩 Summary

Create a new way to set up a development environment based only on containers and with support for [devcontainers](https://containers.dev).

### 🔗 Related issues (if any)

Closes #570
Closes #566

### 💬 Other information

There is more to do in terms of documentation. We need to rewrite [Getting started web](https://cpu.dsek.se/doc/getting-started-VImh85Ngwa) if this is merged.


.devcontainer/compose.yaml
```diff
@@ -0,0 +1,36 @@
+services:
```
Tomas-Ka, line 1: isn't this missing our meilisearch and poppler services? I feel like it could be useful to have those even in a dev environment.

davidagardh: I intentionally left them out, but we can discuss if that's the right choice. In my opinion, most common development tasks don't need them. I'm choosing to prioritize keeping the default setup small over having a production-like development environment. Our current setup is also a compromise. It doesn't include scheduler-service or shlink in the default dev workflow and minio requires extra configuration.

Tomas-Ka: Reasonable, but I do believe the search function is highly useful when debugging mostly anything, since it allows a user to easily traverse the webpage, and make simple queries as to what data is loaded (Members, Positions and so on). With that said Meilisearch requires to be seeded when started though which I don't believe most people do, so it may be unnecessary as you say.

davidagardh: I added a second configuration that includes search, but without is default.

.devcontainer/search/devcontainer.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "D-sek web (with search)",
+  "dockerComposeFile": "../compose.yaml",
+  "service": "web",
+  "runServices": ["web", "db", "meilisearch", "poppler-server"],
+  "workspaceFolder": "/workspace",
+  "forwardPorts": [5173, 5555, 7700],
+  "onCreateCommand": "yes | pnpm install && pnpm migrate && pnpm seed",
```
Tomas-Ka, line 8: Something here is going wrong. When starting the container with `podman compose up` and navigating to localhost:5173 I get the following error:

Image

Tomas-Ka (changes requested):

.devcontainer/search/devcontainer.json
Tomas-Ka: Web can connect to the database, but it's empty

Tomas-Ka: neevermind, this is part of the dev container

Tomas-Ka (approved):
Haven't tested the devcontainer integration, but the image itself works great

memagu:
Works om my machine! VSC on plain Arclonüx

davidagardh:
I want to more work on this before merging. It currently doesn't work if using podman as the container backend for devcontainers in VSCode. I also want to better support some more advanced workflows, like connecting to an external database. Thanks to everyone who tested it! Expect a new PR this summer.

PR #1192, opened 2026-05-17
Add a script && workflow to check translation synchronisation

### LINKED ISSUE #1187: CI to check that translations exist in both English and Swedish
### 🧩 Problem statement

Currently Swedish and English translation files miss lines that exist in the other translation.

### 💡 Proposed solution

Create a pre-commit hook that double checks that all translation strings exist in both files.

### 📍 Affected area(s)

CI / CD - translation files.

### ✅ Acceptance criteria

- [ ] When I make a commit I can be assured that I'm not missing any translations in any of my translation files

### 🗳️ Alternatives considered

_No response_

### ❓ Next steps/open questions

Should this be done user side or on github CI?
I am leaning towards user side (as I would personally like it to always be checked), but it's worth asking.

SLUCHABLUB:
### 🧩 Summary

Adds a script && workflow to check translation synchronisation

### 🔗 Related issues (if any)

Closes #1187 

### 💬 Other information

This adds a script to the `dev` directory. This made writing the workflow easier and as a side effect made the check easier to do locally.

I also removed the translation strings that only existed in one language. I do not think this will pose a problem since they were unusable anyway.


SLUCHABLUB:
> since they were unusable anyway

i forgot to account for the fact that we're using javascript

dev/check_translations.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+tmp="$(mktemp -d)"
+trap 'rm -rf "$tmp"' EXIT
```
Tomas-Ka, line 6: This is genius, and I can't believe I've never seen it before. Will be stealing

src/translations/sv.json
```diff
@@ -938,10 +938,6 @@
   "nav_nollning_description": "Ett sammanhang inom vilket du kan lära känna studentlivet och nya människor",
   "nav_about_guild": "Om sektionen",
   "nav_about_guild_desc": "Sektionen består av 15 utskott med olika ansvarsområden.",
-  "nav_news_feed": "Nyhetsflöde",
-  "nav_news_feed_desc": "Se de senaste nyheterna på D-sektionen.",
```
Tomas-Ka, line 942: This is used in src/routes/routes.ts:69, and should probably instead of getting removed have an English translation.

src/translations/sv.json
```diff
@@ -938,10 +938,6 @@
...
-  "nav_events": "Evenemang",
-  "nav_events_desc": "Se aktuella evenemang på D-sektionen.",
```
Tomas-Ka, line 944: This is used in src/routes/routes.ts:76, and should probably instead of getting removed have an English translation.

src/translations/sv.json
```diff
@@ -1001,7 +997,6 @@
   "fileupload_error": "Sidan tillåter inte nedladdning av filer.",
   "datetimeselector_range_error": "Intervall slutar innan det börjar ",
   "timepicker_invalid_time": " Ogiltig tid",
-  "datepicker_pick_date": "Pick a date",
```
Tomas-Ka, line 1004: This is used in src/lib/components/datetime-selector/DatePicker.svelte:70, and is also in English. This line should be moved to the English file, and another entry with the Swedish translation should be added here.

src/translations/sv.json
```diff
@@ -1040,7 +1035,6 @@
   "committees_card_image": "Kortbild (i utskottslitan)",
   "committees_banner_image": "Bannerbild",
   "committees_invert_text": "Invertera text/logo",
-  "committees_edit_links": "Redigera länkar",
```
Tomas-Ka, line 1043: This is used in src/routes/(app)/committees/LinksDialog.svelte:37 and 26 and should probably instead of getting removed have an English translation.

Tomas-Ka (changes requested):
Script looks great. Just a few things about the removed translation files.
Also, if you could add an entry to the package.json to allow running the translation check with `pnpm translation-check` or similar that would be great
For example by adding this:
`    "translation-check": "./dev/check_translations.sh"
This could be added to the `ci-check` action as well, in which case we wouldn't need a separate workflow for it as it would already run in the existing one.

Tomas-Ka:
well fuck
bad timing

src/translations/sv.json
SLUCHABLUB: Fixed

src/translations/sv.json
SLUCHABLUB: Fixed

src/translations/sv.json
SLUCHABLUB: Fixed

src/translations/sv.json
SLUCHABLUB: Fixed

SLUCHABLUB:
I'll add it to `package.json` as well.

SLUCHABLUB:
I've added it to `package.json` and moved the check into the existing CI workflow. I guess I thought it would be cleaner with a separate workflow but in hindsight, this is probably a lot better.

package.json
Tomas-Ka: we should probably add pnpm:check-translations to this list as well (and potentially pnpm:test since it seems to be missing).
This is what I use to run ci checks locally, so I don't forget anything before I push to remote, and I would find it very nice to have it in here.

dev/check_translations.sh
```diff
@@ -0,0 +1,26 @@
...
+jq --raw-output 'keys[]' src/translations/sv.json >"$tmp/sv-keys"
+
+if cmp --quiet "$tmp/en-keys" "$tmp/sv-keys"; then
+    exit 0
+fi
+
+only_en_keys="$(comm -23 "$tmp/en-keys" "$tmp/sv-keys")"
+only_sv_keys="$(comm -13 "$tmp/en-keys" "$tmp/sv-keys")"
+
+if [[ -n "$only_en_keys" ]]; then
```
Tomas-Ka, line 18: A great QOL feature would be to grep after the non-complete keys in the codebase to give the user a starting point when looking for what the value is used for.

Tomas-Ka: This is certainly not required to merge though

Tomas-Ka (changes requested):
I would like to have the new check-translations action in the general pnpm action for CI. After looking more closely at it I think it's reasonable to let this be it's own worker (instead of bundling it with everything else). I mostly pushed to consolidate since I believed the workflow used the `pnpm ci-check` command already

SLUCHABLUB:
I did not add `pnpm build` since it was wack and seems to mutate files (I'm not totally sure what it "checks").

package.json
SLUCHABLUB: Fixed

dev/check_translations.sh
SLUCHABLUB: I'm not sure what you're getting at. I think that if this lint is ever to fire, it will be because the developer edited the strings, in which case they will know what the strings are used for.

Tomas-Ka (approved):
this looks great, tyvm!

PR #1204, opened 2026-06-03
New design: /applying

fgren:
### 🧩 Summary
Adds the /applying page (sökande), which is new. It has information about the three programmes, and about the student life.


src/translations/sv.json
```diff
@@ -933,8 +933,8 @@
   "positions_years": "år",
   "positions_history": "Historik",
   "applicant": "Sökande",
-  "nollning_nav_picture_title": "Nollningen 2025",
-  "nollning_theme": "En nollning från himlavalvet",
+  "nollning_nav_picture_title": "Nollningen 2026",
+  "nollning_theme": "Ny student? Klicka här!",
```
Tomas-Ka: this should probably have another name to not confuse future devs

src/translations/en.json
```diff
@@ -933,8 +933,8 @@
   "positions_years": "years",
   "positions_history": "History",
   "applicant": "Applicant",
-  "nollning_nav_picture_title": "Introduction 2025",
-  "nollning_theme": "An Introduction from the Celistial Sphere",
+  "nollning_nav_picture_title": "Introduction 26",
+  "nollning_theme": "New student? Click here!",
```
Tomas-Ka: this should probably have another name to not confuse future devs

src/translations/en.json
```diff
@@ -1096,5 +1096,21 @@
...
+  "applying_student": "Being a student",
+  "applying_student_desc": "Do you wonder about what it's like studying at the university? On the Lund University website you can find more information about what it is like to study and being a student.",
+  "applying_student_cta": "Study at Lund University (lu.se)",
+  "applying_student_link": "https://www.lunduniversity.lu.se/study/study-lund-university",
+  "applying_housing": "Housing",
+  "applying_housing_desc": "Do you want to know more about finding housing in Lund? On Studentlund's website, you'll find information about different types of housing and housing queues..",
```
Tomas-Ka: remove double periods

Tomas-Ka (changes requested):
some small nitpicks, haven't checked design, but code looks ok, so I assume it's good

Isak-Kallini:
Some pictures here would be nice?
image

Maybe pictures for each program section as well?

Tomas-Ka (changes requested):
Code and design look good (although I have begun to realize the text on the site might be a bit too small overall, not a problem for this commit however).
Only one big thing to edit, and that is the "för Studenter"/"for Students" button on the homepage, which still redirects to /sokande.

I also wonder if there's a reason the links at the bottom of the applying page are all purple instead of a mix of all our primary colours, or a neutral fourth option?

fgren:
I've changed the link on the landing page, but I don't think it should point towards this page. "For students" seems aimed at current students, but since we don't have a "Become a volunteer" page or an event calendar yet, this'll do.
Also changed up the button colors slightly. Making the third button pistachio would make sense, but it doesn't really fit in well with rosa and lila when they're all used side-by-side.

Tomas-Ka:
That sounds reasonable!

Tomas-Ka (approved):
Looks good to me!

PR #1206, opened 2026-06-05
fix admin access

fgren:
### 🧩 Summary
replaces the auth check on the access pages with `CREATE` instead of `UPDATE`, since no one has update (because updating access policies isn't supported, only creating and deleting them`

### DIFF
--- a/src/routes/(app)/admin/access/[apiName]/+page.server.ts
+++ b/src/routes/(app)/admin/access/[apiName]/+page.server.ts
@@ -33,7 +33,7 @@ export type DeleteSchema = Infer<typeof deleteSchema>;
 
 export const load: PageServerLoad = async ({ locals, params }) => {
   const { prisma, user } = locals;
-  authorize(apiNames.ACCESS_POLICY.UPDATE, user);
+  authorize(apiNames.ACCESS_POLICY.CREATE, user);
 
   const policies = await prisma.accessPolicy.findMany({
     where: {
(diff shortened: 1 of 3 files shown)


src/routes/(app)/admin/access/[apiName]/+page.server.ts
```diff
@@ -33,7 +33,7 @@ export type DeleteSchema = Infer<typeof deleteSchema>;
 
 export const load: PageServerLoad = async ({ locals, params }) => {
   const { prisma, user } = locals;
-  authorize(apiNames.ACCESS_POLICY.UPDATE, user);
+  authorize(apiNames.ACCESS_POLICY.CREATE, user);
```
Comment (Tomas-Ka) [inline src/routes/(app)/admin/access/[apiName]/+page.server.ts:36]: I'd say the read access policy is probably a better fit for being able to view the policies, but this works as well.

src/routes/(app)/admin/access/positions/+page.server.ts
```diff
@@ -15,7 +15,7 @@ const createPolicySchema = z.object({
 export const load: PageServerLoad = async ({ locals }) => {
   const { prisma, user } = locals;
 
-  authorize(apiNames.ACCESS_POLICY.UPDATE, user);
```
Tomas-Ka, line 18: See other comment

Tomas-Ka (approved):
I think read fits better than create, since this is for viewing the entries. This is serviceable though

src/routes/(app)/admin/access/[apiName]/+page.server.ts
Comment (fgren) [reply src/routes/(app)/admin/access/[apiName]/+page.server.ts:36]: The downside to using read is that everyone (*) has read, since you should be able to see your own access policies.

Comment (Tomas-Ka) [reply src/routes/(app)/admin/access/[apiName]/+page.server.ts:36]: Ah that makes sense

PR #1210, opened 2026-07-04
fix news scheduler

Isak-Kallini:
### 🧩 Summary

fixes new scheduler by adding frontend

### 🔗 Related issues (if any)

Closes #

### DIFF
--- a/src/routes/(app)/news/+page.server.ts
+++ b/src/routes/(app)/news/+page.server.ts
@@ -8,9 +8,14 @@ import {
   getPageOrThrowSvelteError,
   getPageSizeOrThrowSvelteError,
 } from "$lib/utils/url.server";
+import { isAuthorized } from "$lib/utils/authorization";
+import apiNames from "$lib/utils/apiNames";
 
 export const load: PageServerLoad = async ({ locals, url }) => {
-  const { prisma, member } = locals;
+  const { prisma, member, user } = locals;
+
+  // News admins can see everyone's scheduled articles, not just their own.
+  const canSeeAllScheduled = isAuthorized(apiNames.NEWS.UPDATE, user);
 
   const articleCount = await prisma.article.count();
   const pageSize = getPageSizeOrThrowSvelteError(url);
@@ -36,11 +41,15 @@ export const load: PageServerLoad = async ({ locals, url }) => {
           publishedAt: {
             gt: new Date(),
           },
-          author: {
-            member: {
-              id: member.id,
-            },
-          },
+          ...(canSeeAllScheduled
+            ? {}
+            : {
+                author: {
+                  member: {
+                    id: member.id,
+                  },
+                },
+              }),
         },
         include: {
           author: {
--- a/src/routes/(app)/news/+page.server.ts
+++ b/src/routes/(app)/news/+page.server.ts
@@ -42,6 +42,23 @@ export const load: PageServerLoad = async ({ locals, url }) => {
             },
           },
         },
+        include: {
+          author: {
+            include: {
+              member: true,
+              mandate: {
+                include: {
+                  position: true,
+                },
+              },
+              customAuthor: true,
+            },
+          },
+          tags: true,
+        },
+        orderBy: {
+          publishedAt: "asc",
+        },
       })
     : [];
 
--- a/src/routes/(app)/news/+page.svelte
+++ b/src/routes/(app)/news/+page.svelte
@@ -8,6 +8,7 @@
   import ArticleCard from "$lib/components/ArticleCard.svelte";
   import { isAuthorized } from "$lib/utils/authorization";
   import apiNames from "$lib/utils/apiNames";
+  import dayjs from "dayjs";
 
   let { data } = $props();
 </script>
@@ -31,6 +32,27 @@
     <a href="/news/create"><Button>+ {m.news_create()}</Button></a>
   {/if}
 </div>
+
+{#if data.scheduledArticles.length > 0}
+  <section class="mb-6">
+    <h2 class="mb-3 text-lg font-semibold">{m.news_scheduledNews()}</h2>
+    <div class="flex flex-col gap-2">
+      {#each data.scheduledArticles as article (article.id)}
+        <a
+          href="/news/{article.slug}/edit"
+          class="hover:bg-muted/50 flex items-center justify-between rounded-md border px-4 py-3 text-sm transition-colors"
+        >
+          <span class="font-medium">{article.headerSv}</span>
+          <span class="text-muted-foreground ml-4 shrink-0">
+            {m.news_scheduledFor()}
+            {dayjs(article.publishedAt).format("YYYY-MM-DD HH:mm")}
+          </span>
+        </a>
+      {/each}
+    </div>
+  </section>
+{/if}
+
 <Pagination pageCount={data.pageCount} class="pb-2" />
 <div class="space-y-4">
   <section class="grid grid-cols-1 gap-8 md:grid-cols-2">
--- a/src/routes/(app)/news/+page.svelte
+++ b/src/routes/(app)/news/+page.svelte
@@ -42,7 +42,7 @@
           href="/news/{article.slug}/edit"
           class="hover:bg-muted/50 flex items-center justify-between rounded-md border px-4 py-3 text-sm transition-colors"
         >
-          <span class="font-medium">{article.headerSv}</span>
+          <span class="font-medium">{article.header}</span>
           <span class="text-muted-foreground ml-4 shrink-0">
             {m.news_scheduledFor()}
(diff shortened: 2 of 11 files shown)


src/routes/(app)/news/create/+page.svelte
```diff
@@ -30,6 +71,13 @@
   committees={data.committees}
 >
   {#snippet formEnd()}
+    <div class="flex w-full flex-col gap-1.5">
+      <Label>{m.news_schedule_publish_time()}</Label>
+      <div class="flex flex-row items-center gap-2">
+        <DatePicker bind:value={publishDate} />
+        <TimePicker bind:value={publishTimeValue} />
+      </div>
+    </div>
```
danieladugyan: There's no way to clear the date field if I accidentally input something. Maybe the input fields shoud even be hidden by a "Schedule" toggle?

Image

src/routes/(app)/news/+page.svelte
```diff
@@ -31,6 +32,27 @@
...
+{#if data.scheduledArticles.length > 0}
+  <section class="mb-6">
+    <h2 class="mb-3 text-lg font-semibold">{m.news_scheduledNews()}</h2>
+    <div class="flex flex-col gap-2">
+      {#each data.scheduledArticles as article (article.id)}
+        <a
+          href="/news/{article.slug}/edit"
+          class="hover:bg-muted/50 flex items-center justify-between rounded-md border px-4 py-3 text-sm transition-colors"
+        >
+          <span class="font-medium">{article.headerSv}</span>
```
danieladugyan:
```suggestion
          <span class="font-medium">{article.header}</span>
```

src/routes/(app)/news/[slug]/edit/+page.server.ts
Comment (danieladugyan) [inline src/routes/(app)/news/[slug]/edit/+page.server.ts:94]: ### AI review
**Scheduling Update Logic Bug**: In `src/lib/news/server/actions.ts`  ( `updateArticle`):
- Old Article Data Used for Scheduling: The scheduled notification logic was running before updating the database, and it was passing the old un-edited  `article`  from the database directly into `scheduleExecution`  or the  `PATCH`  request. This would have caused the notification to send out the old article title/content even after you updated it.
- Broken Condition for Initially Drafted Articles: The scheduling logic was guarded by  `if (article?.publishedAt > new Date() ...)`  using the old article's  `publishedAt`  time. If an article was originally saved with a past/empty publish time (like a draft) and you later edited it to schedule it in the future, it would completely bypass the  `if`  block, failing to schedule the notification.
- Fixes: Refactor the logic to  `prisma.article.update`  the article first, and then use the new updated values and the new  `publishTime`  to determine whether to create, update, or cancel the scheduled task.
[actions.ts](https://github.com/user-attachments/files/29675761/actions.txt)

src/routes/(app)/news/create/+page.svelte
danieladugyan, line 1: Can't test creating a news article locally, it crashes with an error.

Image

Image

danieladugyan (changes requested):

src/routes/(app)/news/+page.server.ts
danieladugyan: ### AI review
Admin Visibility: Currently, news administrators cannot see other users' scheduled articles on the main news page.

src/routes/(app)/news/create/+page.svelte
Isak-Kallini: Did you turn on the scheduler service locally?

danieladugyan: i guess not? 🤷, did `docker compose up -d`  and skimmed the docs but didn't find anything else

Isak-Kallini: hmm, `docker compose up` in `./scheduler-service` should be enough :thinking:

src/routes/(app)/news/[slug]/edit/+page.svelte
Comment (danieladugyan) [inline src/routes/(app)/news/[slug]/edit/+page.svelte:1]: suggestion: I think this file (and the corresponding create route) could have a bit more idiomatic Svelte 5 code.

> Effects are an escape hatch and should mostly be avoided. In particular, avoid updating state inside effects. [...]
> - If you need to run some code in response to user interaction, put the code directly in an event handler or use a function binding as appropriate.

I've provided some in-line code suggestions in separate comments.

src/routes/(app)/news/[slug]/edit/+page.svelte
```diff
@@ -1,13 +1,114 @@
...
+  let publishDate = $state<string | undefined>(
+    $form.publishTime
+      ? toCalendarDate(fromDate($form.publishTime, tz)).toString()
+      : undefined,
+  );
+  let publishTimeValue = $state<Time>(
+    $form.publishTime
+      ? toTime(fromDate($form.publishTime, tz))
+      : parseTime("12:00"),
+  );
```
Comment (danieladugyan) [inline src/routes/(app)/news/[slug]/edit/+page.svelte]:
```suggestion
  // These are immutable objects that are only re-assigned,
  // so using $state.raw avoids overhead.
  let publishDate = $state.raw<string | undefined>(
    $form.publishTime
      ? toCalendarDate(fromDate($form.publishTime, tz)).toString()
      : undefined,
  );
  let publishTimeValue = $state.raw<Time>(
    $form.publishTime
      ? toTime(fromDate($form.publishTime, tz))
      : parseTime("12:00"),
  );

```
Probably not something we'll ever notice, but I suppose it's a neat touch.

src/routes/(app)/news/[slug]/edit/+page.svelte
Comment (danieladugyan) [reply src/routes/(app)/news/[slug]/edit/+page.svelte:1]: suggestion: While we're at it, I reckon these two routes could avoid duplication by just placing all of the code that's currently in the `formEnd` snippet directly inside ArticleEditor (along with all of the schedule state logic). Apart from a type cast, they are identical right now.

src/routes/(app)/news/[slug]/edit/+page.svelte
```diff
@@ -1,13 +1,114 @@
...
+        parsed.month,
+        parsed.day,
+        publishTimeValue.hour,
+        publishTimeValue.minute,
+      );
+      $form.publishTime = dt.toDate(tz);
+    } else {
+      $form.publishTime = null;
+    }
+  });
```
Comment (danieladugyan) [inline src/routes/(app)/news/[slug]/edit/+page.svelte]:
```suggestion
  function updatePublishTime() {
    if (schedulePublish && publishDate) {
      const parsed = parseDate(publishDate);
      const dt = new CalendarDateTime(
        parsed.year,
        parsed.month,
        parsed.day,
        publishTimeValue.hour,
        publishTimeValue.minute,
      );
      $form.publishTime = dt.toDate(tz);
    } else {
      $form.publishTime = null;
    }
  }
```
This function can be called when the form fields are updated.

src/routes/(app)/news/[slug]/edit/+page.svelte
```diff
@@ -1,13 +1,114 @@
...
+>
+  {#snippet formEnd()}
+    <div class="flex w-full flex-col gap-1.5">
+      <div class="flex flex-row items-center gap-2">
+        <Checkbox
+          bind:checked={schedulePublish}
+          id="schedulePublish"
+          class="p-2"
+          onCheckedChange={(checked) => {
+            if (!checked) publishDate = undefined;
```
Comment (danieladugyan) [inline src/routes/(app)/news/[slug]/edit/+page.svelte]:
```suggestion
            if (!checked) publishDate = undefined;
            updatePublishTime();
```

src/routes/(app)/news/[slug]/edit/+page.svelte
```diff
@@ -1,13 +1,114 @@
...
+          }}
+        />
+        <Label for="schedulePublish">{m.news_schedule_publish_time()}</Label>
+      </div>
+      {#if schedulePublish}
+        <div class="flex flex-row items-center gap-2">
+          <DatePicker bind:value={publishDate} />
+          <TimePicker bind:value={publishTimeValue} />
```
Comment (danieladugyan) [inline src/routes/(app)/news/[slug]/edit/+page.svelte]:
```suggestion
          <DatePicker
            bind:value={() => publishDate,
            (value) => {
              publishDate = value;
              updatePublishTime();
            }}
          />
          <TimePicker
            bind:value={() => publishTimeValue,
            (value) => {
              publishTimeValue = value;
              updatePublishTime();
            }}
          />
```
This uses [function bindings](https://svelte.dev/docs/svelte/bind#Function-bindings) to perform additional logic when the values are updated.

src/routes/(app)/news/create/+page.svelte
danieladugyan: Nice! 👍

src/routes/(app)/news/[slug]/edit/+page.server.ts
Comment (danieladugyan) [reply src/routes/(app)/news/[slug]/edit/+page.server.ts:94]: I think it's fine now? It's quite intricate 😅

danieladugyan (approved):
Nice, LGTM! 👍

PR #1216, opened 2026-07-06
General fixes

### LINKED ISSUE #1217: Committee positions are sorted incorrectly on mobile

fgren:
### 🧩 Summary
Attempt to fix the sorting on the committee page (/about)
Fix document sorting (backport of existing logic)
Update contact page with new DELTA logo
Improve contact page
Fix committee position sorting (closes #1217)
Make /documents responsive
Fix navbar access for members


Isak-Kallini:
I have already fixed some of these, may have pushed to main :sweat_smile:

src/routes/(app)/committees/[shortName]/members/+page.svelte
```diff
@@ -33,8 +33,8 @@
       />
     {/each}
   </div>
-  <div class="flex w-full flex-col gap-4 lg:hidden">
-    {#each data.committee.positions as position, index (position.id)}
+  <div class="flex w-full flex-col gap-4 lg:hidden" id="test">
```
Comment (Isak-Kallini) [inline src/routes/(app)/committees/[shortName]/members/+page.svelte]: why id="test"?

src/routes/(app)/Navbar.svelte
```diff
@@ -41,7 +41,9 @@
     $props();
 
   const canAccess = (accessRequired: string | null) =>
-    accessRequired === null || isAuthorized(accessRequired, page.data.user);
+    accessRequired == "_"
```
Isak-Kallini: this should not be needed. When you are logged in you should have the `_` access policy. It looks like that, and `*` does not exist in prod for some reason.
I think it's better to fix the real issue rather than a quick fix like this

Isak-Kallini: I just realized that i am wrong, `_` and `*` are roles, not access polices. But I still think we should find some other solution than this fix, we should not use `_` as an access policy

src/routes/(app)/committees/[shortName]/members/+page.svelte
Comment (fgren) [reply src/routes/(app)/committees/[shortName]/members/+page.svelte]: left over from testing

src/routes/(app)/Navbar.svelte
fgren: Agree that another solution would be preferred, even though this is the only edge case (since * corresponds to null). It would probably be better to use an access policy such as `core:member`?

Isak-Kallini: sounds good

Isak-Kallini (approved):
fix access, otherwise lgtm

PR #1219, opened 2026-07-07
Remove data-sveltekit-reload attribute from <body>

danieladugyan:
### 🧩 Summary

Remove the `data-sveltekit-reload` attribute from the `<body>` tag to enable client-side routing.

### 💬 Other information

Maybe this is intentional and needed during the migration?

### DIFF
--- a/src/app.html
+++ b/src/app.html
@@ -14,11 +14,7 @@
     />
     %sveltekit.head%
   </head>
-  <body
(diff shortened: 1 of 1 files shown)


Isak-Kallini:
This is intended, otherwise things break when you navigate from a new design page to one that reroutes to the old.
It should be removed once everything is migrated though

Issue #1222, opened 2026-07-07
The Fate of the Footer


christofergartner:
My solution to #1215 is to remove the study bank and trivia links. However, I feel that a decision ought to be made on if these should be permanently axed, or if they could one day make a glorious return.

Are the study bank and trivia of interest?

PR #1223, opened 2026-07-07
Clean up the footer

### LINKED ISSUE #1215: Fix links in footer
Some of them link to incorrect pages

christofergartner:
The initial intention was to fix some broken links, but in the end I made some further (primarily semantic) changes:
* The Taube quote is now a `<blockquote>`, as opposed to an `<h1>` (more semantically correct, and, in my opinion, more aesthetically pleasing
* Make the `<aside>` into a plain `<div>` (it was hard to justify calling it an "aside", since it does not specifically relate to the main content of any page where it is included)
* Modify the link hrefs
* Remove the study bank and trivia anchors
* Make the entire footer wrapped in a `<footer>` rather than a plain `<div>`

### 🔗 Related issues (if any)

Closes #1215, but open #1222 (in a sense).

### 💬 Other information

Thanks to @fgren and @RufusKoggRojder for feedback and help!


fgren (commented):
Looks good! The only thing I'm skeptical about is the appearance of the blockquote. On shorter pages, it stands out a lot. I'm thinking that something like `font-medium tracking-tight not-italic` would be better, since it doesn't grab as much attention (but keeps the left line, inset and semantics).

christofergartner:
@fgren I redid the quote with inspiration from what you said. I tried a few variations on style, and am happy with the result. I used tighter tracking and a less weighted font.

I will die on the hill of retaining italic, because I feel there is no soul left without it, so I will keep it.

Are you happy with this? :)

danieladugyan:
> I will die on the hill of retaining italic, because I feel there is no soul left without it, so I will keep it.

Sorry 😅, as always I think I might prefer the current boring one. I'm fine either way though so I'll leave you two to battle over this one 😄

fgren (approved):
I'm fine with keeping the quote in italics, but I added the correct font so it uses the italic version of IBM Plex Sans instead of tilting the standard font. (is more correct and looks better, in my opinion)

PR #1226, opened 2026-07-08
Header semantics

christofergartner:
### 🧩 Summary

Rename the Navbar to Header and adapt its semantic elements.


src/routes/+page.svelte
christofergartner, line 1: I did not remove anything. The large diff is because of adjusted indentation, wrapping the former header in main.

src/routes/+page.svelte
```diff
@@ -80,121 +80,124 @@
...
-  <div class="absolute top-1/4 px-10 pl-6 md:top-1/3 md:pl-10 lg:pl-44">
-    <h1
-      class="font-sans text-5xl font-bold break-keep text-white uppercase sm:mb-10 sm:text-7xl lg:text-9xl xl:text-[125px]"
-    >
-      {m.dsektionen()}
-    </h1>
-    <p class="mb-10 max-w-prose font-medium text-white lg:mb-9 lg:text-xl">
-      {m.landing_intro()}
-    </p>
+<div class="contents">
```
christofergartner, line 83: Looks strange but I didn't dare remove the `<div>` without understanding its function. (The site looks identical for me without it, however).

danieladugyan: Hm, I can't that figure out either. I know [it's needed in app.html](https://svelte.dev/docs/kit/project-structure#:~:text=svelte%3Ahead%3E%20content-,%25sveltekit.body%25,-%E2%80%94%20the%20markup%20for), but [it seems pointless](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/display#contents) here. I suppose we'd probably break something down the line if we removed it so might as well keep it around.

src/routes/+page.svelte
```diff
@@ -80,121 +80,124 @@
...
-      <Button href="/applying" size="lg" class="text-lg">
-        <GraduationCap class="size-6" />{m.landing_forStudents()}
-      </Button>
-      <Button
-        href="/info/for-foretag"
-        size="lg"
-        class="hover:bg-secondary-background/30 text-lg text-white"
-        variant="outline"
+<main>
+  <section
```
danieladugyan: nitpick: I'd probably keep this as a `<header>` instead of a `<section>`.
> The `<header>` HTML element represents introductory content, typically a group of introductory or navigational aids. It may contain some heading elements but also a logo, a search form, an author name, and other elements.
— [`<header>` HTML header element - HTML | MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/header)

danieladugyan (approved):
Minor nitpick, LGTM! 👍

christofergartner:
@danieladugyan I applied your suggestion and will merge it now. Thanks for the help! :)

PR #1229, opened 2026-07-10
Fix read permissions for stock routes. Change preview script from pm2 to

stagrim:
should work

pls look and merge 🥺️

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "ci-check": "concurrently 'pnpm:lint' 'pnpm:format' 'pnpm:check' --prefix-colors 'auto' --group",
     "dev": "vite dev",
     "build": "vite build",
-    "preview": "pm2 start prod/ecosystem.config.cjs --no-daemon",
+    "preview": "vite preview",
     "lint": "eslint .",
     "lint:fix": "eslint . --fix",
     "format": "prettier . --check",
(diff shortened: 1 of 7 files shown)


Isak-Kallini:
Add the check to +layout.server.ts instead so it runs for all /admin/stocklist routes without having to add it in many places?

Isak-Kallini (approved):

danieladugyan (changes requested):
> Add the check to +layout.server.ts instead so it runs for all /admin/stocklist routes without having to add it in many places?

That's a common footgun in SvelteKit that should be avoided[^1] (see https://svelte.dev/docs/kit/load#Implications-for-authentication).

There are a few alternatives. You could revert the latest commit, but the proper solution would be to change the `read` permission in `schema.zmodel` for `DrinkItem` from `@@allow("read", true)` to `@@allow("read", has(auth().policies, "drinkitem:read"))`.

Long-term if we'd like to move away from ZenStack in favour of route level access, the recommended approach seems to be having a `Map<Route, RequiredPermission>` that's checked in `hooks.server.ts`.

[^1]: *Why* it should be avoided is another question. FWIW I _think_ we could just about get away with it here, but it will encourage people to use the same approach in places where it actually ends up causing an issue.

PR #1273, opened 2026-08-26
positions fixes + add nollning

fgren:
### 🧩 Summary
- Removes old positions page that wasn't implemented
- Adds a year selector to the committee positions page
- Adds the nollning theme


src/routes/(app)/positions/+page.svelte
Isak-Kallini, line 1: Do we not want this page anymore? I think it was kind of nice to have an overview of all positions, it is also difficult to find positions that don't exist anymore without this page

Isak-Kallini (approved):

src/routes/(app)/positions/+page.svelte
fgren: I deleted the wrong page

PR #1282, opened 2026-09-02
the big fix :)

Fiery-132:


backend/WHY_GO_BACKEND.md
```diff
@@ -0,0 +1,421 @@
...
+    .startOf("year")
+    .add(weekNum - 1, "week");
+}
+```
+
+Its `updateSchedule` action goes further — zod validation, three `isAuthorized` calls interleaved with business rules, and a raw `prisma.cafeShift.findMany`, all in one function. There's no layer you could call "the cafe service" or "the cafe handler" — SvelteKit has no opinion about separating those, and nothing here introduced one. Section 2 shows the replacement: five files, five jobs.
+
+### 1.2 A generic key-value table standing in for a real feature
+
+"When does nollning start/end" used to live in a generic `AdminSetting` table, hand-parsed, behind a module-level 1-hour cache:
```
Isak-Kallini, line 34: This is not a real argument for go, we can easily fix it without go

backend/WHY_GO_BACKEND.md
```diff
@@ -0,0 +1,421 @@
...
+ }
+ return phaseFor(season), nil
+}
+```
+
+`Current`/`Phase`/`NollaYear`/`InjectStabenPolicy` are four documented, independently testable methods. No cache, no string parsing — `GetCurrentSeason` is an indexed query against a real table.
+
+### 1.3 Authorization reimplemented at the call site, over and over
+
+`isAuthorized(...)` is called **27 times across 18 files** (`grep -rn "isAuthorized(" src`), each a hand-written check re-derived wherever a route happens to need it — three separate times in `cafe/+page.server.ts` alone, none of them the same "can this user edit this shift" logic reused. There's no service method that owns the answer; it's scattered across whatever `.svelte`/`.server.ts` files needed a check. That count is from the _current_ tree, seven phases into the port — it's what's left after every already-ported feature moved its own call sites to Go's `auth.Require`, not the original size of the problem.
```
Isak-Kallini, line 96: I think `isAuthorized` is not really used for security in our codebase, it's rather used for business logic/ui rendering and I think that is a valid use case. We have zenstack or `authorize` for security.

backend/WHY_GO_BACKEND.md
```diff
@@ -0,0 +1,421 @@
...
+
+Some of this reimplementation happened **client-side**, computing the same decision the server separately re-checks — and it drifted: both the article and event detail pages had `canEdit`/`canDelete` hardcoded `true`, because reimplementing the check in TypeScript was never finished (`DESIGN.md`'s Principle #5 was written in direct response to this). And it's not just UI gating — mandate CRUD and governing-document create/update had **no server-side authorization at all**, relying solely on ZenStack's `@@allow` (1.4) as a backstop. None of this was exotic; nothing structurally required a mutation to declare its policy.
```
Isak-Kallini, line 98: Relying on zenstack is intentional, it is our primary tool for access control so I don't think this is valid.

backend/WHY_GO_BACKEND.md
```diff
@@ -0,0 +1,421 @@
...
+
+**Now:** every huma route runs through `auth.Middleware` once; a handler declares its policy requirement and the framework enforces it before the body runs — one call, one place, checked against a real `Identity` (policy strings are centralized in `internal/apinames`, section 5). Mandate and governing-document mutations are gated explicitly now; `canEdit`/`canDelete` are resolved server-side and just read by the frontend (section 7). Nollning's `InjectStabenPolicy`/`NollaYear` (1.2) follow the same discipline one level up, so no downstream domain reimplements "is this a nollning period" itself.
```
Isak-Kallini, line 100: Will only ever using the middleware be enough? I am thinking about the cafe example above:
```js
if (!isAuthorized(apiNames.CAFE.SEE_ALL_WEEKS, user)) {
  if (weekNum < currentWeek.week() || weekNum > currentWeek.week() + 2) {
    error(403, { message: m.cafe_error_no_week_viewing_perms() });
  }
}
```
Would this check not be needed in go as well?

backend/WHY_GO_BACKEND.md
```diff
@@ -0,0 +1,421 @@
...
+    @@allow("create", has(auth().policies, "news:article:create"))
+    @@allow("read", has(auth().policies, "news:article:manage"))
+    @@allow("read", auth().memberId == article.authorId)
+    @@allow("update", has(auth().policies, "news:article:manage"))
+    @@allow("delete", auth().memberId == article.authorId)
+    @@allow("delete", has(auth().policies, "news:article:manage"))
+}
+```
+
+Declarative-looking, but it means the whole app's authorization surface lives in one 1,365-line non-Go/TS DSL, enforced by a runtime Prisma wrapper (`enhance()`) that silently filters queries — you find out by an empty result, not a `WHERE` clause you can read. And `@@allow` isn't even the only place authorization is decided (1.3): two overlapping systems, no single source of truth. `auth.Require` (1.3) is that single source now.
```
Isak-Kallini, line 118: After thinking about it I'm less negative about zenstack actually. While it can be a little unintuitive to understand the first time it still achieves one valuable thing: all access control in one place. While we have `authorize` and `isAuthorized`, these are more convenience functions imo. They are there to make the ui look nicer, authorize to block for example admin pages and isAuthorized to hide edit buttons etc. Zenstack is still the source of truth in the end, these functions are just there to hide buttons you can't use or block a page that would be empty for you anyways. But they are never vital to actual security.

I think we should discuss zenstack more irl

backend/WHY_GO_BACKEND.md
```diff
@@ -0,0 +1,421 @@
...
+
+### 1.5 A god-file `hooks.server.ts`
+
+279 lines run on every request today — smaller than it used to be, since session/identity logic has already started moving to Go (see "Now," below); the original file was bigger and did even more in one place. What's still there: identity resolution, member bootstrapping, an anonymous-user cookie, a ZenStack-`enhance()`d Prisma client, an onboarding redirect buried in the middle, plus the ZenStack RPC mount, CORS, and metrics — one function doing all of it regardless of which feature the request is for.
+
+**Now:** `auth.Middleware` resolves identity, `locale.Middleware` resolves locale, and nothing else does either job. The SvelteKit side shrank to asking Go "who is this" and building `locals` from the answer.
```
Isak-Kallini, line 124: This just feels like splitting it up. All the code is still there, it just looks a bit nicer. I don't think this is a valid argument for go, it just moves responsibility.

backend/WHY_GO_BACKEND.md
```diff
@@ -0,0 +1,421 @@
...
+
+### 1.6 Stringly-typed logic producing silent bugs
+
+- The old "minutes" resolver matched `["Protokoll", "Minutes, Minute"]` — a comma inside one string, so `"Minutes"`/`"Minute"` alone never matched anything. **Now:** three real substrings.
+- Every document-delete action always deleted from the `documents` bucket, silently broken for SRD/requirement files that live in `files`. **Now:** the bucket is derived from the document type server-side.
+- Gallery uploads pushed promises into a list that was never `await`ed. **Now:** `POST /gallery/upload` awaits every file.
+
+None of these are exotic — they're what happens when logic is re-derived by hand at each call site instead of computed once, with no checkpoint to catch it.
```
Isak-Kallini, line 132: I'm not sure I see how go would prevent these bugs

backend/WHY_GO_BACKEND.md
```diff
@@ -0,0 +1,421 @@
...
+
+### 1.7 Two features that look like one, because they share a table
+
+`Markdown` backs both generic `/info` pages and each committee's about/links text; a separate `Document` model (governing documents) happens to share the `/documents` URL prefix with an unrelated MinIO file-browsing feature. Nothing about the naming distinguishes them — a natural result of a codebase with no domain boundaries. Go's package-per-domain naming (`internal/governingdocs` vs. `internal/markdown`) can't collide this way; huma refuses to register two packages under the same name at all.
```
Isak-Kallini, line 136: I'm not sure I understand this one, please explain irl sometime

backend/WHY_GO_BACKEND.md
```diff
@@ -0,0 +1,421 @@
...
+}, func(ctx context.Context, input *songSlugInput) (*songOutput, error) {
+ song, err := svc.Get(ctx, input.Slug)
+ if err != nil {
+  return nil, humaServiceError(err)
+ }
+ return &songOutput{Body: *song}, nil
+})
+```
+
+Five files, five jobs, and none of them know how to do each other's: the query doesn't know HTTP exists, the handler doesn't know SQL exists, `toSong` is the only place a database row and an API response ever touch. Swap "song" for "alert," "event," or any other ported domain and the file names change; the shape doesn't.
```
Isak-Kallini, line 241: This just reads like a convention, nothing in go actually enforces this and it relies solely on the programmer following these rules.

backend/WHY_GO_BACKEND.md
```diff
@@ -0,0 +1,421 @@
...
+
+## 3. Why this needs to be a real API, not just a cleanup pass
+
+**A native app is coming, and it can't run SvelteKit's server code.** Per `DESIGN.md`: "a mobile app will call it too, going through none of SvelteKit's code at all — so any validation or authorization logic living in TypeScript is either redundant, since Go has to enforce it for real regardless, or a second, drifting copy of a rule the mobile app never sees." A rule that lives only in a `+page.server.ts` action is invisible to that app by construction — it has to live somewhere a non-browser client can trigger and rely on.
+
+**The coupling in section 1 is a security liability, not a style complaint.** Missing-auth bugs (1.3) and dual authorization systems (1.4) are exactly what "add more discipline" doesn't reliably fix, because nothing requires a mutation to declare its policy. An API layer where a handler doesn't compile or route without one closes that gap by construction, not by review vigilance.
```
Isak-Kallini, line 348: This is not a valid argument for go, we can achieve this in sveltekit as well

backend/WHY_GO_BACKEND.md
```diff
@@ -0,0 +1,421 @@
...
+
+**The realistic SvelteKit alternative isn't `+page.server.ts` as-is — it's remote functions as a shared service layer.** Business logic moves into `.remote.ts` functions (SvelteKit's analogue of `service.go`); the website calls them directly; a thin `+server.ts` wraps the same functions in JSON for the app. A real, coherent design — that still loses to Go:
+
+- **Two calling conventions, permanently.** The website calls logic in-process; the app calls the same logic through an HTTP wrapper — forever, not just during migration. Go has one interaction surface, used identically by both. Porting a feature to the app becomes "copy the API call the website already makes," not "write a new wrapper."
+- **The wrapper doesn't reliably stay thin.** Several old `helpers.ts` files started as thin call-forwarders and grew real logic — this codebase has already broken that promise more than once.
+- **Two call sites for the same logic is section 1.3's drift pattern, prospectively.** A check added to the wrapper and forgotten on the direct path (or vice versa) is exactly how `canEdit`/`canDelete` disagreed. One contract used identically by both clients has no second path to drift from.
```
Isak-Kallini, line 358: I don't think this has to be a problem. I think having a thin wrapper for the api is a reasonable convention that is at least as easy, if not easier to maintain than the suggested "5 layers" convention suggested in this doc.

backend/WHY_GO_BACKEND.md
```diff
@@ -0,0 +1,421 @@
...
+- **It doesn't fix Prisma/ZenStack underneath** — the same `enhance()`-wrapped client and fused schema+policy file (1.4) either way.
+- **Still shares the frontend's runtime and dependency graph** — one Node process, one `node_modules`, no independent scaling or deploy.
+
+**What Go actually buys**, independent of the native-app question:
+
+- **`internal/` is a compiler-enforced boundary; `.server.ts` is a naming convention.** The Go toolchain refuses to let outside code import `internal/` — a build error, not a lint rule. SvelteKit's equivalent is enforced by a Vite plugin at bundle time, invisible to `svelte-check`/SSR checks — and the identical mistake (a `.server.ts`-only import leaking into a universal `+page.ts` load) was made three separate times during this very migration, by people actively restructuring the code. A boundary that fragile isn't one a rotating volunteer base can rely on.
```
Isak-Kallini, line 364: This is kind of valid, but I think the advantage is overstated. Specifically `/internal` doesn't really do much here, it only enforces boundaries within go. ts would never import anything from go anyways and it's rather the separation into a different service that is the win here. Nothing stops internal/api/huma_songs.go from importing internal/db and running SQL directly.

backend/WHY_GO_BACKEND.md
```diff
@@ -0,0 +1,421 @@
...
+- **No shared module graph with the frontend, period.** No `package.json`, `node_modules`, or TypeScript program in common — there's no `import` statement, in either direction, that could cross the boundary by accident, because there's no shared build system to cross through.
+- **A smaller stack per change.** Safely touching `cafe/+page.server.ts` needs SvelteKit's load/actions model, Prisma, ZenStack's `@@allow`, superforms+zod, dayjs plugins, and the `apiNames` convention — six things, one file. A Go domain package needs three, each in its own file, following a template a dozen-plus packages already demonstrate.
+- **huma makes the typed contract the API.** The OpenAPI spec is derived from Go's own types at routing time — it can't drift the way a hand-maintained spec can. `openapi-typescript`/`openapi-fetch` generate the TS side straight from it (section 2's frontend snippet).
```
Isak-Kallini, line 367: We can achieve this in ts as well

backend/WHY_GO_BACKEND.md
```diff
@@ -0,0 +1,421 @@
...
+
+- **`internal/` is a compiler-enforced boundary; `.server.ts` is a naming convention.** The Go toolchain refuses to let outside code import `internal/` — a build error, not a lint rule. SvelteKit's equivalent is enforced by a Vite plugin at bundle time, invisible to `svelte-check`/SSR checks — and the identical mistake (a `.server.ts`-only import leaking into a universal `+page.ts` load) was made three separate times during this very migration, by people actively restructuring the code. A boundary that fragile isn't one a rotating volunteer base can rely on.
+- **No shared module graph with the frontend, period.** No `package.json`, `node_modules`, or TypeScript program in common — there's no `import` statement, in either direction, that could cross the boundary by accident, because there's no shared build system to cross through.
+- **A smaller stack per change.** Safely touching `cafe/+page.server.ts` needs SvelteKit's load/actions model, Prisma, ZenStack's `@@allow`, superforms+zod, dayjs plugins, and the `apiNames` convention — six things, one file. A Go domain package needs three, each in its own file, following a template a dozen-plus packages already demonstrate.
```
Isak-Kallini, line 366: I think the spirit of this codebase is that you should not have to worry about the parts that you are not touching. I feel like the go solution puts more trust in the developer which I'm not sure is a win considering one of our goals is to be beginner friendly.

backend/WHY_GO_BACKEND.md
```diff
@@ -0,0 +1,421 @@
...
+- **huma makes the typed contract the API.** The OpenAPI spec is derived from Go's own types at routing time — it can't drift the way a hand-maintained spec can. `openapi-typescript`/`openapi-fetch` generate the TS side straight from it (section 2's frontend snippet).
+- **sqlc makes SQL bugs a compile error**, not a silently-empty result — the opacity problem is ZenStack's policy layer specifically, since Prisma's own queries are typed.
```
Isak-Kallini, line 368: Not sure what is meant with sql bugs? Zenstacks access control works as intended is a different question entirely and I don't think this is a fair comparison

backend/WHY_GO_BACKEND.md
```diff
@@ -0,0 +1,421 @@
...
+
+## 7. The frontend gets simpler, not just the backend
+
+Once a domain is ported, SvelteKit collapses to "call the generated client, render the result":
+
+- `client.ts` + `schema.d.ts`, generated from Go's own `/openapi.json`, are the _entire_ integration surface — no hand-written DTOs or fetch wrapper, full autocomplete straight from the Go structs.
+- Every translated field ships both the raw pair and one resolved field computed server-side — display code just reads it. (No more `translationExtension.ts` - TACK OCH LOV 🎉)
```
Isak-Kallini, line 402: I don't think the way translations are handled in this pr is an improvement. It's basically some helper functions that you have to use at every instance where there are translations and basically do `lang === "en" ? encontent : svcontent`. We could both do this in sveltekit if we want, but I think the translationExtension is a much more elegant solution.

Isak-Kallini (commented):
I think in general the document attributes a lot of things as an advantage to go when it really is just an advantage of the rewrite and can just as well be done in sveltekit.
I think the biggest wins for go are the backend/fronend boundary and that it is an independent, scalable service.

Some wins for sveltekit that wasn't mentioned:
- we get end-to-end types for free within the website
- very easy to migrate incrementally
- we can reuse the good things that we already have

However I think the most important point is what we think about zenstack. After thinking about it I don't think that it's obvious that removing it is the correct choice. It would mean porting the whole zenstack.zmodel with row-level rules with nothing that can really guarantee that we got it right. I think this is a decision we should not make too lightly and I'm not as sure that zenstack is that bad anymore.

I hope I don't come of as too negative, I think the document touches on a lot of important points and is nice to base discussion on.

backend/WHY_GO_BACKEND.md
Fiery-132: Fair, but it would involve a massive restructuring and, if we are already willing to go that far, why not take the opportunity to switch backend?

backend/WHY_GO_BACKEND.md
Fiery-132: That is valid, but I still think it is a random helper applied inconsistently. And even if it's not a security concern, we still need it for UX.

backend/WHY_GO_BACKEND.md
Fiery-132: I agree with your point - zenstack is currently our source of access control and this is written kind of badly. But I think there is an argument to make here for duplication of auth logic. We often need to apply the same checks both client side and server side because buttons should be conditionally hidden and then the server-side action of the same button should be gated and throw errors upon missing access. And under all of this is zenstack just quietly returning nothing upon missing access policies, which is what forces us to handle all the UX at every call site.

backend/WHY_GO_BACKEND.md
Fiery-132: True, you would need to authorize it. I think it just becomes easier for the developer to do so. The shape will always be the same for every request, and I feel it is easier to reason about "should this endpoint be authorised?", rather than sprinkling it among all the other logic inside a `+page.server.ts` file.

backend/WHY_GO_BACKEND.md
Fiery-132: I agree we should discuss it irl. However, I'll leave a response here too. :)

After thinking about, I'm more negative about zenstack actually, haha. I personally think it will always be unintuitive because it just silently returns nothing if you are missing permissions. On top of this, it's a huge file that can be very intimidating and hard to navigate, and groups the logic separated from the actual call sites. Lastly, like you mentioned, we also have `authorize` and `isAuthorized` that should be used whenever zenstack has access policies on data we want to use. So since we already have all the policies and double-check them at every use, we could just do without zenstack, provided the auth functions are used consistently. This makes auth more explicit and easier to reason about, I think, by keeping the rules next to the actual code.

On top of this, we also have many uses of `authorizedPrismaClient`, which shows we already feel the need to bypass zenstack. In go, this would not be an issue since we only apply auth where we actually need it.

backend/WHY_GO_BACKEND.md
Fiery-132: It's not a perfect argument for go in the sense that this *could* be accomplished in ts, too. But I feel the nature of how you construct services and handlers in go promotes this design. The same would be true if we were working with an express server in ts, for example. And I think having the code "look nicer" is actually a genuine benefit. If it's easy to look at, it will be easy to maintain. For example, keeping separate middlewares means there is no risk for any one of them to leak into the next and start sharing responsibilities.

backend/WHY_GO_BACKEND.md
Fiery-132: I agree this is pretty weak (chapter 1 overall is weakest imo) but the main argument is in the last sentence here.
> None of these are exotic — they're what happens when logic is re-derived by hand at each call site instead of computed once, with no checkpoint to catch it.
But yes, an API in sveltekit would essentially accomplish the same thing.

backend/WHY_GO_BACKEND.md
Fiery-132: I suppose. But there is value in convention. Consider on-boarding, it is significantly easier to understand if every place follows the same pattern. Then it just becomes a matter of copying this pattern and adding your own custom logic inside the service file.

backend/WHY_GO_BACKEND.md
Fiery-132: Fair. This section is mostly justifying the need for an actual API. However, it would require a huge rewrite to fix this while keeping our current code base.

backend/WHY_GO_BACKEND.md
Fiery-132: Hm, I disagree. We have been pretty bad at this historically - helper files just grow and grow for example. Things are not refactored, but instead just piled on top of one another. But even a perfectly-thin wrapper is still something someone writes and maintains by hand, including its JSON shape. There is no OpenAPI-from-remote-functions generator. While we can generate this in go, derived directly from the declared types. This means sveltekit would require more manual code duplication.

To add to this, using remote functions means that the app and website consume data differently, which leads to even more unique code to maintain. Maintainability of the native app is already a concern so is it not beneficial for the app and website to consume the same API? Then, porting features almost becomes "copy the API call and write a frontend to represent it". And an eventual change in the API will require the same change in both places. Not a change to the wrapper and then two different changes, one for web and one for app.

backend/WHY_GO_BACKEND.md
Fiery-132: Sure, we rely mostly on convention to stop that from happening. But we agree here
> ts would never import anything from go anyways and it's rather the separation into a different service that is the win here

and I think this is the most important part. This permanently prevents any leakage and sharing between server-side and client-side functions.

backend/WHY_GO_BACKEND.md
Fiery-132: Maybe, but I don't really see the vision.

I have already responded to the last point, regarding wrappers and duplication.

No shared frontend and backend is something we legitimately can not enforce in sveltekit. 

"A smaller stack per change" is maybe something we could accomplish but not by simply moving load and action functions to remote function files. If we design a good way to actually break up all this then maybe. But it's true that the way code is written today, you need extensive knowledge of the codebase to change even small things because they are so tightly coupled. (SvelteKit's load/actions model, Prisma, ZenStack's `@@allow`, superforms+zod, dayjs plugins, and the `apiNames` convention inside a single function)

backend/WHY_GO_BACKEND.md
Fiery-132: I'm not sure i understand. In my mind this allows the developer to worry less about the parts they aren't touching. The most obvious example is that when touching the frontend you don't se any of the backend logic. But even on the backend you don't need to worry about the db interaction when working on the actual api endpoint, for example. I also feel that the more explicit, and less "magical", things are the easier they will be to understand and work with.

Besides, I feel like this allows for new ways to introduce people to the code. You could easily task a beginner with fixing a small thing on the frontend without confusing them by passing data back and forth between two or three different `page`/`layout` files, not to mention they don't even need to know about the database.

backend/WHY_GO_BACKEND.md
Fiery-132: Yeah I also feel this bullet sort of mixes two concerns. 

It is a plus that SQL bugs are compile time bugs because then all your queries are checked. But prisma also kind of does this with its generated types. Although that is hijacked by translationExtension and you need to import a special client and special types. But that's a bit of a sidetrack.

The part about zenstack is the "silently-empty result". I responded to this earlier already but I think it's better if a failed auth is loud and obvious, rather than just filtering out data behind the scenes.

backend/WHY_GO_BACKEND.md
Fiery-132: Hm, I don't really agree and I think it sounds worse than it actually is. I wrote much of translationExtension and I don't actually think it is that elegant. First of all, I barely understand it which is a big minus whenever it breaks next time. And because we override prisma, we can't really point to the docs for how to use it so you need to remember to import a custom type and then index it via strings to get the correct return type instead of just using prismas generated types.

The second part of this is that it is actually quite elegant even in go. Take this snippet for example:
```go
Alert{
	ID:        dbutil.UUIDStr(row.ID),
	Severity:  row.Severity,
	Message:   dbutil.ResolveName(row.MessageSv, &row.MessageEn, loc),
	MessageSv: row.MessageSv,
	MessageEn: row.MessageEn,
	CreatedAt: row.CreatedAt.Time,
}
```
We already need to construct the return type so it's only natural to resolve the language here too. Explicit so easy to understand and hard to get wrong.

Fiery-132:
That was a lot of valid feedback and I tried to respond as well as I could to all of it. We will definitely need to discuss this in person, though. And I agree that our view on zenstack is an important discussion.

To reply to your final points though:
> we get end-to-end types for free within the website

True. But also true for go since we can generate types for the frontend. So it's almost the same thing: types declared only in one place.

> very easy to migrate incrementally

This one is a valid concern. I just think the benefits are worth the tradeoff.

> we can reuse the good things that we already have

True. I'm aware I might be a bit negative so it's hard to see the actual good things we do have. But even so, we can still reuse much of it because the sveltekit project will not disappear. It's just that parts of it will move out.

Also hope my responses are not too dismissive or negative. I do think you brought up a lot of valid points. I'm just still convinced go is the ideal way forward. 😅

PR #1283, opened 2026-09-05
Fix songbook for the greater good

stagrim:
### 🧩 Summary

Create a separate songbook page which is meant to mirror the latest edition of the songbook. Simpler search that only filters text of song titles, and numbers by page. The song archive is left in an awkward spot here, but I suggest leaving it like this, until a new "archive" page can make it obsolete.

image
song book interface. UI borrowed from the song archive.
image

Add link to nav bar

### 💬 Other information

Still kind of rushed, but I think something like this should be in place during the novice period since the current song archive is very difficult to search. 

### Next steps

Improve the song archive search algorithm and make it usable again. Then both can complement each other.


danieladugyan (commented):
LGTM perhaps? Too much code 😅

I have some spontaneous thoughts:
- I don't think this should be two different pages – it's a bit confusing to have two pages that are so similar. I think what's implemented here is way more useful than what we currently have, and the archive songs can just be an add-on IMO.
- `SongSearch` has quite a bit of generic logic, feels like `debounce` and `isFetching` should be something we can implement once and reuse.
- The database model looks fine to me (apart from some `camelCase` columns that should be `snake_case`), but I'm not sure `numberOnPage` is needed.
- I would keep the `/songbook` URL and perhaps the page title too.

Thoughts? Maybe @Fiery-132 @Isak-Kallini @fgren?

Steindt (approved):
Lgtm

PR #1294, opened 2026-09-09
Fix iOS notification badge not clearing

Fiery-132:
## Summary
- `AppUnreadNotificationHandler` and `AppNotificationTokenHandler` were dropped from `(app)/+layout.svelte` when the old design was removed in 5b797ca9, silently severing the bridge that tells the native wrapper app the unread notification count (used to set/clear the iOS badge) and that uploads push tokens.
- Android masked this because its shade badge clears on notification dismissal independent of app code; iOS's badge is a persistent counter that only updates via the now-unreachable `setBadgeCountAsync` call.
- Re-mounts both handlers, gated on `data.isApp`, matching how the old layout wired them up.

## Test plan
- [x] `pnpm check` passes with no errors
- [x] `eslint` passes on the changed file
- [ ] Manual verification on an iOS device that opening notifications clears the app badge

### DIFF
--- a/src/routes/(app)/+layout.svelte
+++ b/src/routes/(app)/+layout.svelte
@@ -3,6 +3,8 @@
   import Alert from "$lib/components/Alert.svelte";
   import Header from "./Header.svelte";
   import Footer from "./Footer.svelte";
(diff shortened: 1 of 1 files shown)


davidagardh (approved):
hope it works

PR #1318, opened 2026-09-20
[Hotfix] Populate edit form, making editing possible

stagrim:
Makes song edit form populated with existing data. It is impossible to edit a song right now since the id field doesn't get populated.


stagrim:
Whoops, missed that part. Seems like awaiting the parent solves the issue then since the form data isn't used. Because the issue definitely exists on main, and is resolved here... Will create a cleaner fix.

PR #1321, opened 2026-09-21
Nuke expenses and stocklist

### LINKED ISSUE #1276: Remove expenses and stocklist
Since we have ekonomi.dsek.se now /expenses and /stocklist is no longer needed.
- Delete all files related to these pages
- Contact skattmästare to make sure they don't need the data in the database
- Remove all data models relating to expenses or stocklist in schema.zmodel
- Make sure that daisy is still working since it has it's own database model (hopefully you don't have to do anything, the /expenses and /stocklist pages crashing on daisy is fine)

SLUCHABLUB:
### 🔗 Related issues (if any)

Closes #1276 

### 💬 Other information

I am yet to check with the treasurer if they want the data left in the DB.


SLUCHABLUB:
The treasurer has said that there are some expenses in the old system that have been made by people in the introductions committee. We either need to convert them to use their legal names, or we need to wait until the end of the introduction. I think the latter is easier since it's only one week away.

PR #1322, opened 2026-09-21
fix recurring events dst

### LINKED ISSUE #881: Add consistency DST for reoccuring events
### Description

**What is happening?**
Reoccuring events do not adjust according to DST, resulting in an hour+- on the event calendar. 

Image

**What is expected to happen:**
Reoccuring events always happen at the same time the first event was published, regardless of DST changing. 

### Steps to reproduce

dsek.se/events/calendar

Look at the reoccurring events when a time shift in DST occurs, e.g. Sunday October 26th 3 am

### Further information

_No response_

Isak-Kallini:
### 🧩 Summary

Why the earlier fix (#971) didn't work: it built each occurrence by taking dayjs.tz("YYYY-MM-DD") (midnight) and calling .hour(h) on it. That object keeps midnight's UTC offset. On the day the clocks change, midnight and the event's hour have different offsets, so the event comes out an hour off. I reproduced it: under TZ=UTC, a daily 18:00 event on Sun 25 Oct 2026 was created at 17:00 Stockholm time. It only appeared on the transition day and depended on the server's timezone, which fits it looking fixed in testing but still failing in production.

The fix:
- src/lib/events/recurrence.ts (new): works on calendar dates and wall-clock times, and only converts to an exact time at the end, by parsing "YYYY-MM-DD HH:mm" in Europe/Stockholm. It has two functions: generateOccurrences and retimeOccurrence.
- src/lib/events/server/actions.ts: createEvent and updateEvent both use those two functions. updateEvent had the same midnight-plus-.hour() pattern.
- Overnight events: the old code also rebuilt the end time from the start date, so an event running 22:00 to 02:00 got an end before its start. The end now stays the same number of calendar days after the start.

Verification:
- Tests: recurrence.test.ts checks the Stockholm wall-clock times using Intl, not dayjs. It covers weekly and daily events across both DST changes, overnight events, and the separation count. It passes with the process timezone set to UTC, Stockholm, Los Angeles and Auckland.
- Old approach: with the old logic swapped in under TZ=UTC, 3 of the 9 tests fail, all on the transition-day cases.
- Lint and types: eslint and the type check are clean for src/lib/events.

Not covered:
- Existing events: recurring events already in the database were created with the old logic, so those with wrong times (like the ones in the issue screenshot) stay wrong. They need a one-off correction or re-creation. I haven't touched any data.
- Create/edit form: the form was removed in "Remove old design", so I could only test the server side, not the form in a browser. The form should send an ISO instant, and the fix assumes that.

### 🔗 Related issues (if any)

Closes #881


davidagardh (approved):
👍

